PR #43, opened 2017-05-08
Add analyzer for Assert.Equal<string> instead of the string-specific overload

ErikSchierboom:
This is my first attempt at an xUnit analyzer, so I'd welcome any feedback!

Refs https://github.com/xunit/xunit/issues/1257


bradwilson:
Looks good to me, but I'm still pretty new to the project. @marcind?

src/xunit.analyzers/AssertEqualStringValueShouldNotUseGenericType.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+        }
+
+        protected override void Analyze(SyntaxNodeAnalysisContext context, InvocationExpressionSyntax invocation, IMethodSymbol method)
+        {
+            var arguments = invocation.ArgumentList.Arguments;
+            if (arguments.Count != 2)
+                return;
+
+            var stringType = context.SemanticModel.Compilation.GetTypeByMetadataName(typeof(string).FullName);
```
marcind: Use `compilation.GetSpecialType(SpecialType.System_String)` or maybe even the IsSpecialType method in the if statement instead of looking the type up at all.

src/xunit.analyzers/AssertEqualStringValueShouldNotUseGenericType.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Xunit.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class AssertEqualStringValueShouldNotUseGenericType : AssertUsageAnalyzerBase
+    {
+        private static readonly HashSet<string> EqualMethods = new HashSet<string>(new[] { "Equal" });
```
marcind: Would it make sense to include StrictEqual as well?

src/xunit.analyzers/AssertEqualStringValueShouldNotUseGenericTypeFixer.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+
+namespace Xunit.Analyzers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public class AssertEqualStringValueShouldNotUseGenericTypeFixer : CodeFixProvider
+    {
+        const string title = "Remove generic type";
```
marcind: Technically you are removing the generic type argument, but the effect here is to use the string-specific Assert.Equal method overload. So maybe "Use string Assert.Equal"?

src/xunit.analyzers/Constants.cs
```diff
@@ -126,6 +126,11 @@ internal static class Descriptors
                 Categories.Assertions, DiagnosticSeverity.Warning, isEnabledByDefault: true,
                 description: "The value type will be boxed which means its identity will always be different.");
 
+            internal static DiagnosticDescriptor X2006_AssertEqualStringValueShouldNotUseGenericType { get; } = new DiagnosticDescriptor("xUnit2006",
+                "Do not use generic type in string equality test",
```
marcind: "Do not use generic Assert.Equal overload for strings" or something like that.

src/xunit.analyzers/AssertEqualStringValueShouldNotUseGenericType.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Xunit.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class AssertEqualStringValueShouldNotUseGenericType : AssertUsageAnalyzerBase
```
marcind: AssertEqualGenericShouldNotBeUsedForStringValue

test/xunit.analyzers.tests/AssertEqualStringValueShouldNotUseGenericTypeTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Xunit.Analyzers
+{
+    public class AssertEqualStringValueShouldNotUseGenericTypeTests
+    {
+        public class Analyzer
+        {
+            readonly DiagnosticAnalyzer analyzer = new AssertEqualStringValueShouldNotUseGenericType();
+
+            public static TheoryData<string, string> Foo { get; } = new TheoryData<string, string>
```
marcind: Data?

marcind (changes requested):
Looks great, except for my nitpicking.

test/xunit.analyzers.tests/AssertEqualStringValueShouldNotUseGenericTypeTests.cs
ErikSchierboom: Good point. I just copied it from an existing tests file.

src/xunit.analyzers/Constants.cs
ErikSchierboom: Sounds much better. I'll correct it.

src/xunit.analyzers/AssertEqualStringValueShouldNotUseGenericTypeFixer.cs
ErikSchierboom: Will do.

src/xunit.analyzers/AssertEqualStringValueShouldNotUseGenericType.cs
ErikSchierboom: Well, the original issue didn't mention it, and there is no string-specific overload of StrictEqual as there is with Equal. Maybe @bradwilson knows?

src/xunit.analyzers/AssertEqualStringValueShouldNotUseGenericType.cs
ErikSchierboom: I will change this.

src/xunit.analyzers/AssertEqualStringValueShouldNotUseGenericType.cs
marcind: The original issues are just idea notes I left myself, frequently incomplete. In this case if there is no behavioral difference between `Equal<string>' and `StrictEqual<string>` then both should be fixed to `Equal(string, string)`

test/xunit.analyzers.tests/AssertEqualStringValueShouldNotUseGenericTypeTests.cs
marcind: Yeah, I know there are a few "placeholder" variable names in the source code. I should fix them.

ErikSchierboom:
@marcind I've updated the PR with your suggested fixes, expect for the StrictEqual comment which I suspect was not what @bradwilson intended.

test/xunit.analyzers.tests/AssertEqualStringValueShouldNotUseGenericTypeTests.cs
ErikSchierboom: @marcind I'll submit a PR :)

bradwilson:
I don't know why someone would choose `StrictEqual` for strings, so I'd tend to want to guide them to the more useful string overloads of `Equal`.

ErikSchierboom:
Would that be a different analyzer then, as this one is targeting the generic overload.

bradwilson:
No opinion. :)

marcind:
Same analyzer, just add StrictEquals to the list of methods checked

bradwilson:
Make sure the Fixer transforms `StrictAssert` into `Assert`, then (I think the existing fixer just strips off the generic).

ErikSchierboom:
@marcind I could add it to the same analyzer, but we should then rename it to something else as translating StrictEqual to Equal does not involve generics. Any ideas?

marcind:
I wouldn't change the name of the analyzer. You'll notice there are a number of other ones that analyze multiple different methods calls (e.g. https://github.com/xunit/xunit.analyzers/blob/master/src/xunit.analyzers/AssertEqualShouldNotBeUsedForBoolLiteralCheck.cs, https://github.com/xunit/xunit.analyzers/blob/master/src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirst.cs, etc).
There's a practical limit to how verbose these names should be. I've been trying to come up with names that describe the general class of methods/concepts being inspected, not enumerating every one of them in the name.

marcind:
Btw, the PRs look good, but we will have to wait a little bit in merging until we finalize the transition of this repo under the .NET Foundation umbrella. It's mostly about covering copyright related issues.  I hope this doesn't temper your enthusiasm.

ErikSchierboom:
@marcind I've updated the PR to also support rewriting `Assert.StrictEqual` and `Assert.StrictEqual<string>` calls. The description displayed has been updated accordingly, although to make them use the same base description, I've had to slightly reword the message. Could you check it the messages are okay?

Good luck with the .NET Foundation transition. It won't deter me at all. I'm having great fun working on these analyzers!

src/xunit.analyzers/AssertEqualGenericShouldNotBeUsedForStringValue.cs
```diff
@@ -0,0 +1,39 @@
...
+            base(Constants.Descriptors.X2006_AssertEqualGenericShouldNotBeUsedForStringValue, EqualMethods)
+        {
+        }
+
+        protected override void Analyze(SyntaxNodeAnalysisContext context, InvocationExpressionSyntax invocation, IMethodSymbol method)
+        {
+            if (invocation.ArgumentList.Arguments.Count != 2)
+                return;
+
+            if (!method.IsGenericMethod && method.Name != "StrictEqual")
```
marcind: I'm not sure the method name check is necessary. [`StrictEqual`](https://github.com/xunit/assert.xunit/blob/master/EqualityAsserts.cs#L90) is generic as well.

marcind (changes requested):

src/xunit.analyzers/AssertEqualGenericShouldNotBeUsedForStringValue.cs
ErikSchierboom: Well yes, but this check is to prevent an analyzer showing for a regular `Assert.Equal` check, which is actually the correct usage.

ErikSchierboom: Having explained that, I think I should change the check to `!method.IsGenericMethod && method.Name == "Equal"`. That better explains what is being verified.

marcind: Makes sense.

marcind (approved):
Waiting for .NET Foundation migration

src/xunit.analyzers/AssertEqualGenericShouldNotBeUsedForStringValue.cs
ErikSchierboom: @marcind I've just updated the PR with the above mentioned change: `method.Name != "StrictEqual"` becomes `method.Name == "Equal"`

marcind:
Closing/reopening for CLA bot

marcind:
Merged in 28ccba2

PR #45, opened 2017-05-08
Add bool regex match to regex match comparison

ErikSchierboom:
Refs https://github.com/xunit/xunit/issues/1261

### DIFF
--- a/src/xunit.analyzers/AssertRegexMatchShouldNotUseBoolLiteralCheckFixer.cs
+++ b/src/xunit.analyzers/AssertRegexMatchShouldNotUseBoolLiteralCheckFixer.cs
@@ -1,0 +1,70 @@
+﻿using System;
+using System.Collections.Immutable;
+using System.Composition;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+
+namespace Xunit.Analyzers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public class AssertRegexMatchShouldNotUseBoolLiteralCheckFixer : CodeFixProvider
+    {
+        const string titleTemplate = "Use Assert.{0}";
+
+        public sealed override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(Constants.Descriptors.X2008_AssertRegexMatchShouldNotUseBoolLiteralCheck.Id);
+
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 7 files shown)


test/xunit.analyzers.tests/CodeAnalyzerHelper.cs
```diff
@@ -15,6 +15,7 @@ class CodeAnalyzerHelper
     {
         static readonly MetadataReference CorlibReference = MetadataReference.CreateFromFile(typeof(object).GetTypeInfo().Assembly.Location);
         static readonly MetadataReference SystemCoreReference = MetadataReference.CreateFromFile(typeof(Enumerable).GetTypeInfo().Assembly.Location);
+        static readonly MetadataReference SystemTextReference = MetadataReference.CreateFromFile(typeof(System.Text.RegularExpressions.Regex).GetTypeInfo().Assembly.Location);
```
marcind, line 18: Is this necessary? The unit tests run on the Desktop framework so this would resolve to System.dll, which should already be included.

src/xunit.analyzers/AssertRegexMatchShouldNotUseBoolLiteralCheckFixer.cs
```diff
@@ -0,0 +1,70 @@
...
+                    invocation
+                        .WithArgumentList(regexIsMatchInvocation.ArgumentList)
+                        .WithExpression(memberAccess.WithName(SyntaxFactory.IdentifierName(replacementMethod))));
+            }
+            else
+            {
+                var regexMemberAccess = (MemberAccessExpressionSyntax)regexIsMatchInvocation.Expression;
+                var regexMember = regexMemberAccess.Expression;
+                
+                editor.ReplaceNode(invocation,
```
marcind, line 60: How robust is this refactoring? E.g. for code like in the unit test FindsWarning_ForInstaceRegexIsMatchWithConstructedRegexVariable, will it do the right thing? What if the regex reference is a parameter, or a field on some other class?

marcind: This makes me wonder if it's worth having this analyzer for the instance method at all. Maybe just the static method?

marcind (changes requested):

src/xunit.analyzers/AssertRegexMatchShouldNotUseBoolLiteralCheckFixer.cs
ErikSchierboom: @marcind Actually, it should also work for those scenarios. I've tested it with two variants:

1. The `Regex` object is _new-ed_ inline
1. The `Regex` object is a variable

I did not test with a parameter, but that should also work. I'll test it tonight. As for the robustness, I think it is good here, as what I am doing is just taking the left part of `x.IsMatch` and using that as the first parameter. Perhaps you could try it on your machine?

test/xunit.analyzers.tests/CodeAnalyzerHelper.cs
ErikSchierboom: It was on my machine, but I'll verify.

src/xunit.analyzers/AssertRegexMatchShouldNotUseBoolLiteralCheckFixer.cs
marcind: Yeah, thinking about this again, instance methods are good. I wasn't thinking straight yesterday.

test/xunit.analyzers.tests/CodeAnalyzerHelper.cs
marcind: I think this is actually a reference to System.dll (at least on desktop), which doesn't look like it was explicitly referenced here yet.

marcind (approved):
Waiting for .NET Foundation transition to complete.

marcind:
Closing/reopening for CLA bot

PR #48, opened 2017-05-08
Use const where possible

ErikSchierboom:


marcind:
Why?

ErikSchierboom:
Well, the most important one is that the [official .NET Design Guidelines](https://msdn.microsoft.com/en-us/library/ms229057(v=vs.110).aspx) recommend using consts over readonly fields: 
> DO use constant fields for constants that will never change.

I also think `const` makes it easier to see that a value cannot be changed, when compared to `static readonly`.

The xUnit source code also appears to use `const` for this purpose: https://github.com/xunit/xunit/blob/914e97d57e388b5c8c2bf5c4381b5c232ed64b5a/src/xunit.runner.utility/Configuration/ConfigReader_Configuration.cs#L101

But feel free to close this one if you think it is unnecessary.

marcind:
Closing/reopening for CLA bot

marcind:
I'm going to close this to focus on merging your other PRs

PR #52, opened 2017-05-15
Fix NRE caused by checking zero-sized arrays in InlineData attributes

natemcmaster:
This fixes a bug in InlineDataMustMatchTheoryParameters which caused a NRE when checking 
```c#
[InlineData(new byte[0])]
```

The tests didn't catch this because `CompilationWithAnalyzers.GetAnalyzerDiagnosticsAsync` does not include `AD0001` which is produced when the analyzer fails (i.e. with NullReferenceException). I've changed the test helper to call `GetAllDiagnosticsAsync` instead.

cc @marcind


marcind:
Closing/reopening for CLA bot

natemcmaster:
Rebased on master.

PR #54, opened 2017-05-31
1298 inline data duplicates

wachulski:
Some early draft. Needs refactoring still and a dedicated fixer.

### DIFF
--- a/src/xunit.analyzers/Descriptors.cs
+++ b/src/xunit.analyzers/Descriptors.cs
@@ -142,6 +142,12 @@ namespace Xunit.Analyzers
             "This includes any combination of static and instance methods declared with any visibility in the same class or across a " +
             "class hierarchy. Rename one of the methods.");
 
+        internal static DiagnosticDescriptor X1025_InlineDataShouldBeUniqueWithinTheory { get; } = new DiagnosticDescriptor("xUnit1025",
+            "InlineData should be unique within the Theory it belongs to",
+            "Theory method '{0}' on test class '{1}' has InlineData duplicate(s).",
+            Categories.Usage, DiagnosticSeverity.Warning, isEnabledByDefault: true,
+            description: "Theory should have all InlineData elements unique. Remove redundant attribute(s) from the theory method.");
+
         internal static DiagnosticDescriptor X2000_AssertEqualLiteralValueShouldBeFirst { get; } = new DiagnosticDescriptor("xUnit2000",
             "Expected value should be first",
             "The literal or constant value {0} should be the first argument in the call to '{1}' in method '{2}' on type '{3}'.",
--- a/src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
+++ b/src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
@@ -1,0 +1,263 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Collections.ObjectModel;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Xunit.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class InlineDataShouldBeUniqueWithinTheory : DiagnosticAnalyzer
+    {
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+            Descriptors.X1025_InlineDataShouldBeUniqueWithinTheory);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterCompilationStartAction(compilationStartContext =>
+            {
+                var compilationTypes = new CompilationTypes(compilationStartContext.Compilation);
+
+                if (!compilationTypes.IsCompilationCapableOfBeingProcessedByThisAnalyzer)
+                    return;
+
+                compilationStartContext.RegisterSymbolAction(
+                    symbolContext => AnalyzeMethod(symbolContext, compilationTypes), SymbolKind.Method);
+            });
+        }
+
+        private static void AnalyzeMethod(SymbolAnalysisContext context, CompilationTypes compilationTypes)
+        {
+            var method = (IMethodSymbol) context.Symbol;
+
+            var methodAllAttributes = method.GetAttributes();
+            if (!methodAllAttributes.ContainsAttributeType(compilationTypes.Theory))
+                return;
+
+            var wellFormedInlineDataAttributes = methodAllAttributes
+                .Where(a => a.AttributeClass == compilationTypes.InlineData
+                            && HasAttributeDeclarationNoCompilationErrors(a, compilationTypes));
+
+            AnalyzeInlineDataAttributesWithinTheory(context, wellFormedInlineDataAttributes);
+        }
+
+        private static void AnalyzeInlineDataAttributesWithinTheory(SymbolAnalysisContext context,
+            IEnumerable<AttributeData> inlineDataAttributes)
+        {
+            var attributeRelatedMethod = (IMethodSymbol)context.Symbol;
+            var uniqueAttributes = new HashSet<AttributeData>(new InlineDataUniquenessComparer(attributeRelatedMethod));
+
+            foreach (var currentInlineData in inlineDataAttributes)
+            {
+                context.CancellationToken.ThrowIfCancellationRequested();
+
+                if (uniqueAttributes.Contains(currentInlineData))
+                {
+                    ReportDuplicate(context, currentInlineData);
+                }
+                else
+                {
+                    uniqueAttributes.Add(currentInlineData);
+                }
+            }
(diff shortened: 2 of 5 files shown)


wachulski:
Well. My mistake. I wanted to create a draft request into my repo master. Nevertheless, treat it as a WIP. TODO still:

- [x] analyzer refactoring
- [x] fixer

You may suggest corrections at this stage (wanted to show the draft early). Questions to the merger for the future:

- where can I discuss technical details (in which direction to design/refactor code)? E.g. for this issue I can see there is some duplication with other analyzer class that can be collapsed into one (utility or sth). Do you have a forum or should I stick to issues discussion like this one?

wachulski:
One more - I plan to rebase it and stash (just to have a single commit for it). So, you'd better not branch from it.

marcind:
We can use this thread for now.

wachulski:
I see no code fix unit tests there yet, thus I'm not going to make a precedence with mine. I consider development ready.

wachulski:
How long does it take to get CLA approved? I signed it a week ago or so. Does it prevent you from reviewing PR?

marcind:
Hi, Sorry for the delay with the review. I've been sick the last few days and am now trying to get caught up with work. Not sure what is happening with the cla bot.

marcind:
Closing and reopening PR to see if it fixes cla bot.

marcind:
As with everything, turn it off and on again fixed the problem.
I'll get to the review once I have a free moment. Likely won't be until this weekend.

wachulski:
Yeah, perfect, thx. Haste makes waste, no problem. I just wanted to make sure it's not deadlocked with wishful thinking both sides.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
```diff
@@ -0,0 +1,154 @@
...
+                    uniquenessMapByHashCodes.Add(argumentsHash, currentInlineData);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Test arguments consist of InlineData arguments concatenated with default parameter values of a test method 
+        /// if such default values are not covered by InlineData arguments.
+        /// </summary>
+        private static IEnumerable<object> GetTestArguments(SymbolAnalysisContext context, AttributeData inlineData)
```
marcind: I prefer IReadOnlyCollection<T> to make multiple lazy enumeration not possible.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
```diff
@@ -0,0 +1,154 @@
...
+            var uniquenessMapByHashCodes = new Dictionary<int, AttributeData>();
+
+            foreach (var currentInlineData in inlineDataAttributes)
+            {
+                context.CancellationToken.ThrowIfCancellationRequested();
+
+                IEnumerable<object> arguments = GetTestArguments(context, currentInlineData);
+                int argumentsHash = GetTestArgumentsAggregatedHashCode(arguments);
+
+                if (uniquenessMapByHashCodes.ContainsKey(argumentsHash))
```
marcind: Just testing the hashes for equality is insufficient and could lead to false positives. You also need to test the two argument collections and see if they are "equal". This could be done by using a Set<AttributeData> with a custom comparer.

(_Technically_, the argument equality check should also consider the corresponding target parameter's type conversion, e.g. `[InlineData(1), InlineData(1L), Theory] void Test(long p) {}` could be marked as having a duplicate even though the arguments themselves are not strictly the same but the test method ends up being called with the same value. **However**, this is probably overkill as dealing with conversions is tricky and it gets further complicated by generic theory methods)

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheoryFixer.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class InlineDataShouldBeUniqueWithinTheoryFixer : CodeFixProvider
+    {
+        public sealed override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(
+            Descriptors.X1025_InlineDataShouldBeUniqueWithinTheory.Id
+        );
+
+        // There is an issue when two (or more) duplicate attributes occur one after another. The batch fixer won't
+        // merge multiple deletes at once due to close vicinity and the fixer would have to execute a few times.
+        // If sb wants all to be deleted at once maybe reporting the original attribute with all its duplicates 
+        // as a collection passed with a fixer parameter. To be considered.
```
marcind, line 24: A `Diagnostic` has an `AdditionalLocations` property that could be used to pass in locations of the other duplicates. But it seems pretty unlikely someone will have multiple duplicates of an InlineData.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheoryFixer.cs
```diff
@@ -0,0 +1,54 @@
...
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var reportedNode = root.FindNode(context.Span);
+            var attributeDuplicate = reportedNode as AttributeSyntax;
+
+            if (attributeDuplicate == null)
+                throw new InvalidOperationException("Expected to have an InlineData attribute duplicate " +
```
marcind: Does this happen often? Under what circumstances? The analyzer always reports a location for an AttributeSyntax so it seems like this condition should not be hit. If it's not hit, remove the check and throw statement?

src/xunit.analyzers/TypeSymbolFactory.cs
```diff
@@ -11,5 +11,20 @@ internal static INamedTypeSymbol IEnumerableOfObjectArray(Compilation compilatio
             var iEnumerableOfObjectArray = iEnumerableOfT.Construct(objectArray);
```
marcind, line 11: Use new method to build objectArray.

marcind (changes requested):

src/xunit.analyzers/Descriptors.cs
```diff
@@ -142,6 +142,12 @@ internal static class Descriptors
             "This includes any combination of static and instance methods declared with any visibility in the same class or across a " +
             "class hierarchy. Rename one of the methods.");
 
+        internal static DiagnosticDescriptor X1025_InlineDataShouldBeUniqueWithinTheory { get; } = new DiagnosticDescriptor("xUnit1025",
+            "InlineData should be unique within the Theory it belongs to",
+            "Theory method '{0}' on test class '{1}' has InlineData that duplicates the one declared at '{2}'.",
```
marcind: Is there some way that these InlineData attributes can be split amongst multiple files? If not, I'd get rid of the logic to print the file name/position and simplify the tests. Typically a collection of inline attributes is easy to scan to find where the duplicate is.

marcind:
Also, please squash into one commit.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
wachulski: good idea, so do I now, thx

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheoryFixer.cs
wachulski: Ok, `AdditionalLocations` might be helpful then. But I agree it's a rare case to see multiple duplicates in one `Theory`. I'd stay with this and eventually handle people's requests after.

src/xunit.analyzers/TypeSymbolFactory.cs
wachulski: It took me a few seconds to work this out, so please confirm: you suggest I put in line 10 in this file the following:
```csharp
var objectArray = GetObjectArrayType(compilation);
```
right?

src/xunit.analyzers/Descriptors.cs
wachulski: Yeah, this adds some complexity with little profit.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
wachulski: I see you point and of course agree.

> This could be done by using a Set with a custom comparer.

Good idea, I'll switch to this.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheoryFixer.cs
wachulski: I know what you mean here. It's a defensive programming check (and yes, it shouldn't be `if-throw` flavor). Any better with: ```Debug.Assert(...)```?

wachulski:
Huh... traversing forests with pattern matching along the way. I should have been more careful when picking low hanging fruits :) Ready for review.

marcind (approved):
Looks great.

PR #55, opened 2017-06-05
Add analyzer to replace Assert.Collection with Assert.Empty when no Action is specified

ErikSchierboom:
Refs https://github.com/xunit/xunit/issues/1256

### DIFF
--- a/src/xunit.analyzers/Descriptors.cs
+++ b/src/xunit.analyzers/Descriptors.cs
@@ -198,5 +198,10 @@ namespace Xunit.Analyzers
             "Do not use boolean check to check for string equality",
             "Do not use {0} to check for string equality.",
             Categories.Assertions, DiagnosticSeverity.Warning, isEnabledByDefault: true);
+
+        internal static DiagnosticDescriptor X2011_AssertEmptyCollectionCheckShouldNotBeUsed { get; } = new DiagnosticDescriptor("xUnit2011",
+            "Do not use empty collection check",
+            "Do not use {0} to check for empty collections.",
+            Categories.Assertions, DiagnosticSeverity.Warning, isEnabledByDefault: true);
     }
 }
--- a/src/xunit.analyzers/AssertEmptyCollectionCheckShouldNotBeUsed.cs
+++ b/src/xunit.analyzers/AssertEmptyCollectionCheckShouldNotBeUsed.cs
@@ -1,0 +1,31 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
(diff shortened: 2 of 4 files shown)


src/xunit.analyzers/Descriptors.cs
```diff
@@ -198,5 +198,10 @@ internal static class Descriptors
             "Do not use boolean check to check for string equality",
             "Do not use {0} to check for string equality.",
             Categories.Assertions, DiagnosticSeverity.Warning, isEnabledByDefault: true);
+
+        internal static DiagnosticDescriptor X2011_AssertCollectionEmptyCheckShouldNotUseCollectionCheckFixer { get; } = new DiagnosticDescriptor("xUnit2011",
```
marcind: Accidental "Fixer" suffix?

src/xunit.analyzers/AssertCollectionEmptyCheckShouldNotUseCollectionCheckFixer.cs
```diff
@@ -0,0 +1,45 @@
...
+        public sealed override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(Descriptors.X2011_AssertCollectionEmptyCheckShouldNotUseCollectionCheckFixer.Id);
+
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var invocation = root.FindNode(context.Span).FirstAncestorOrSelf<InvocationExpressionSyntax>();
+            
+            context.RegisterCodeFix(
```
marcind: I think there should be an alternative fixer for this, which would add an element inspector. If you don't have the bandwidth/interest to implement that we could open a follow up issue instead.

marcind (changes requested):

src/xunit.analyzers/Descriptors.cs
ErikSchierboom: Whoops. I'll fix it

src/xunit.analyzers/AssertCollectionEmptyCheckShouldNotUseCollectionCheckFixer.cs
ErikSchierboom: Excellent suggestion. I'll add it to the PR

ErikSchierboom: @marcind I've added a second code fix to add an element inspector. At the moment, the inspector is just an empty lamdba: `x => { }`. Is that okay or would you like to use something else?

I've also slightly renamed the code fix to not have it tied to the empty collection check.

marcind (approved):

PR #57, opened 2017-06-21
Fixing InlineData(null) handling for InlineData uniqueness analyzer.

wachulski:
Fixes xunit/xunit#1324


marcind:
Does this handle `[InlineData]` as well?

wachulski:
> Does this handle [InlineData] as well?

See the tests related to this change. Do they answer your question?

marcind (approved):

PR #62, opened 2017-07-22
Fix typo in README

jamesqo:
People will try to do `Install-Package xunit.analyzer` and waste time figuring out why it doesn't work, so here's to prevent that.


bradwilson:
Thanks!

PR #63, opened 2017-07-24
Add Assert.Throws/ThrowsAny analyzer

ErikSchierboom:
Refs https://github.com/xunit/xunit/issues/1248

### DIFF
--- a/src/xunit.analyzers/AssertThrowsShouldNotBeUsedForAsyncThrowsCheckFixer.cs
+++ b/src/xunit.analyzers/AssertThrowsShouldNotBeUsedForAsyncThrowsCheckFixer.cs
@@ -1,0 +1,152 @@
+﻿using System;
+using System.Collections.Immutable;
+using System.Composition;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+
+namespace Xunit.Analyzers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public class AssertThrowsShouldNotBeUsedForAsyncThrowsCheckFixer : CodeFixProvider
+    {
+        const string TitleTemplate = "Use Assert.{0}";
+
+        public sealed override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(
+            Descriptors.X2014_AssertThrowsShouldNotBeUsedForAsyncThrowsCheck.Id,
+            Descriptors.X2014_AssertThrowsShouldNotBeUsedForAsyncThrowsCheck_Hidden.Id);
+
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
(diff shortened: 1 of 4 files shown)


marcind:
The generic versions of Assert.Throws have overrides that match a Task-returning method that are marked as obsolete with error, so the following two usages are already marked as compilation errors:

```C#
Assert.Throws<Exception>(() => Foo());
Assert.Throws<ArgumentException>("foo", () => Foo());
```

In those cases the analyzer should not raise an additional warning but just provide a refactoring to fix the problem.

The 3rd handled case `Assert.Throws(typeof(Exception), () => Foo());` does not have an obsoleted overload for some reason, but causes a runtime error thus in this case the analyzer should produce an error.

ErikSchierboom:
Ah good point. How can I best do that?

src/xunit.analyzers/AssertThrowsShouldNotBeUsedForAsyncThrowsCheckFixer.cs
```diff
@@ -0,0 +1,156 @@
...
+                return method.ReturnType;
+
+            var semanticModel = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+            var methodSymbol = semanticModel.GetSymbolInfo(method.ReturnType, cancellationToken).Symbol as ITypeSymbol;
+            var taskType = semanticModel.Compilation.GetTypeByMetadataName(typeof(Task).FullName);
+
+            if (taskType.IsAssignableFrom(methodSymbol))
+                return method.ReturnType;
+
+            return
```
marcind: This implementation builds a return type expression that's fully namespace qualified, even if the document already has the `System.Threading.Tasks` namespace included. I've used `editor.Generator.TypeExpression` to get something takes the imported namespaces into account.

marcind (changes requested):

marcind:
There's `Microsoft.CodeAnalysis.CodeRefactorings.CodeRefactoringProvider`, but I have not explored it that much. It could also probably be done using one descriptor with a severity of hidden (see X2001_AssertEqualsShouldNotBeUsed) and one with an Error.

src/xunit.analyzers/AssertThrowsShouldNotBeUsedForAsyncThrowsCheckFixer.cs
ErikSchierboom: Thanks for the tip! I didn't know how to do that, so that's why I used the fully qualified namespace.

ErikSchierboom:
@marcind I've updated the PR with the following changes:

- The two obsolete methods are not flagged as warnings anymore but use the hidden severity.
- I've added unit tests for the paramName overload (forgot about them).
- I've used `editor.Generator.TypeExpression` to generate the minimal `Task` return type syntax.
- I've rebased on the latest `master`.

src/xunit.analyzers/Descriptors.cs
```diff
@@ -220,6 +220,16 @@ internal static class Descriptors
             "Do not use {0} to check for collection size.",
             Categories.Assertions, DiagnosticSeverity.Warning, isEnabledByDefault: true);
 
+        internal static DiagnosticDescriptor X2014_AssertThrowsShouldNotBeUsedForAsyncThrowsCheck { get; } = new DiagnosticDescriptor("xUnit2014",
+            "Do not use throws check to check for asynchronously thrown exception",
+            "Do not use {0} to check for asynchronously thrown exceptions.",
+            Categories.Assertions, DiagnosticSeverity.Warning, isEnabledByDefault: true);
```
marcind: This should be Error, since not addressing the problem will cause the test to fail at runtime.

marcind (changes requested):

src/xunit.analyzers/Descriptors.cs
ErikSchierboom: Fixed!

ErikSchierboom:
Thanks for the thorough reviews!

PR #64, opened 2017-07-24
Add use generic Assert.Throws overload analyzer

ErikSchierboom:
Refs https://github.com/xunit/xunit/issues/1259


ErikSchierboom:
I noticed that with the introduction of this analyzer, there were two analyzers (IsType and Throws) where the fixer replaced the first `typeof()` parameter with a generic type argument overload. As such, I decided to create a single fixer that can handle both the `IsType` and `Throws` analyzer. If you think this is a bad approach, feel free to let me know.

marcind (approved):

PR #65, opened 2017-07-25
Analyzers for precision parameter of Assert.Equal overload for double…

wachulski:
…s and decimals

Fixes xunit/xunit#1303

### DIFF
--- a/src/xunit.analyzers/AssertEqualPrecisionShoulBeInRangeFixer.cs
+++ b/src/xunit.analyzers/AssertEqualPrecisionShoulBeInRangeFixer.cs
@@ -1,0 +1,52 @@
+﻿using System.Collections.Immutable;
+using System.Composition;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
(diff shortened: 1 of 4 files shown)


src/xunit.analyzers/AssertEqualPrecisionShoulBeInRangeFixer.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+
+namespace Xunit.Analyzers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public class AssertEqualPrecisionShoulBeInRangeFixer : CodeFixProvider
+    {
+        private const string title = "Use precision 0";
```
wachulski, line 17: I'm not sure if it's the best option we can deliver. At least it's a straightforward one and will fail tricky test cases, so it should force people to think what is the best precision for their cases.

marcind (approved):

PR #66, opened 2017-07-30
Warn when users call IsType with an interface/abstract class

jamesqo:
`IsType` verifies that an object is exactly of a given type, which isn't possible for interfaces/abstract classes. I've written an analyzer that detects usages of `Is{Not}Type` with such types and issues a warning.

This PR also contains a code fix that changes `IsType` -> `IsAssignableFrom` (there's no `IsNotAssignableFrom`, so nothing to do for `IsNotType`). I haven't been able to test out the code fix, though, since none of the code fixes seem to work when debugging the VSIX. (**edit:** managed to get around those issues and test out the code fix providers in VS.)

/cc @marcind

### DIFF
--- a/src/xunit.analyzers/AssertIsTypeShouldNotBeUsedForAbstractTypeFixer.cs
+++ b/src/xunit.analyzers/AssertIsTypeShouldNotBeUsedForAbstractTypeFixer.cs
@@ -1,0 +1,51 @@
+﻿using System.Collections.Immutable;
+using System.Composition;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+
+namespace Xunit.Analyzers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public class AssertIsTypeShouldNotBeUsedForAbstractTypeFixer : CodeFixProvider
+    {
+        private const string Title = "Use Assert." + IsAssignableFrom;
+
+        private const string IsAssignableFrom = "IsAssignableFrom";
+
+        private const string IsType = "IsType";
+
+        public sealed override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(Descriptors.X2018_AssertIsTypeShouldNotBeUsedForAbstractType.Id);
+
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var invocation = root.FindNode(context.Span).FirstAncestorOrSelf<InvocationExpressionSyntax>();
+            var methodName = invocation.GetSimpleName()?.Identifier.Text;
+            if (methodName == IsType)
+            {
+                context.RegisterCodeFix(
+                    CodeAction.Create(
+                        Title,
+                        createChangedDocument: ct => UseIsAssignableFromAsync(context.Document, invocation, ct),
+                        equivalenceKey: Title),
+                    context.Diagnostics);
+            }
+        }
+
+        private static async Task<Document> UseIsAssignableFromAsync(Document document, InvocationExpressionSyntax invocation, CancellationToken cancellationToken)
+        {
+            var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
+            var simpleName = invocation.GetSimpleName();
+            editor.ReplaceNode(simpleName, simpleName.WithIdentifier(SyntaxFactory.Identifier(IsAssignableFrom)));
+            return editor.GetChangedDocument();
+        }
+    }
+}
--- a/src/xunit.analyzers/Descriptors.cs
+++ b/src/xunit.analyzers/Descriptors.cs
@@ -262,7 +262,10 @@ namespace Xunit.Analyzers
             "Do not use Contains() to check if a value exists in a collection.",
             Categories.Assertions, DiagnosticSeverity.Warning, isEnabledByDefault: true);
 
-        // Placeholder for rule X2018
+        internal static DiagnosticDescriptor X2018_AssertIsTypeShouldNotBeUsedForAbstractType { get; } = new DiagnosticDescriptor("xUnit2018",
+            "Do not compare an object's exact type to an abstract class or interface",
+            "Do not compare an object's exact type to the {0} '{1}'.",
+            Categories.Assertions, DiagnosticSeverity.Warning, isEnabledByDefault: true);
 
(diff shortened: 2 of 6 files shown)


src/xunit.analyzers/xunit.analyzers.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageTargetFallback>portable-net45+win8</PackageTargetFallback>
     <RootNamespace>Xunit.Analyzers</RootNamespace>
     <TargetFramework>netstandard1.1</TargetFramework>
+    <DebugType>full</DebugType>
```
jamesqo, line 15: @marcind, I was having a really hard time debugging the VSIX because the breakpoints weren't being hit. I eventually realized the fix was to add this, because the portable PDB format doesn't seem to be supported for VSIX projects yet.

jamesqo:
I can't seem to be able to try out any of the code fix providers, including the one in this PR. They work fine when I use a release build from NuGet with `Install-Package xunit.analyzers`, but when I build the repo and start debugging they don't seem to trigger.

You can see here that the analyzer works, but the code fixer doesn't for whatever reason:

cfp

I'm waiting for a response at https://github.com/dotnet/roslyn/issues/21209 to see how I can fix this.

src/xunit.analyzers/Descriptors.cs
```diff
@@ -254,7 +254,10 @@ internal static class Descriptors
             "Keep precision in range {0} when asserting equality of {1} typed actual value.",
             Categories.Assertions, DiagnosticSeverity.Error, isEnabledByDefault: true);
 
-        // Placeholder for rule X2017
+        internal static DiagnosticDescriptor X2017_AssertIsTypeShouldNotBeUsedForAbstractType { get; } = new DiagnosticDescriptor("xUnit2017",
```
marcind: Please rebase and update this rule to have the id `X2018`. Sorry about the conflict :/

src/xunit.analyzers/Descriptors.cs
```diff
@@ -254,7 +254,10 @@ internal static class Descriptors
...
+            "Do not assert that an object's exact type is an abstract class or interface",
+            "Do not assert that an object's exact type is the {0} '{1}'.",
+            Categories.Assertions, DiagnosticSeverity.Error, isEnabledByDefault: true);
```
marcind: Since asserting on an interface or abstract class does not actually cause the test to blow up, this should be a warning, not an error.

src/xunit.analyzers/AssertIsTypeShouldNotBeUsedForAbstractTypeFixer.cs
```diff
@@ -0,0 +1,51 @@
...
+        public sealed override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(Descriptors.X2017_AssertIsTypeShouldNotBeUsedForAbstractType.Id);
+
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var invocation = root.FindNode(context.Span).FirstAncestorOrSelf<InvocationExpressionSyntax>();
+            var methodName = invocation.GetGenericName()?.Identifier.Text;
+            if (methodName == IsType)
```
marcind: I wonder if the best fix for the `IsNotType` cause would be to remove the assertion altogether.

marcind (changes requested):
Some small feedback. Could you also squash when doing the rebase? I prefer to have a clean commit history with one commit per feature. Thanks

src/xunit.analyzers/AssertIsTypeShouldNotBeUsedForAbstractTypeFixer.cs
jamesqo: @marcind Why? That would only confuse the user, who is probably looking for something like `IsNotAssignableFrom` (only that method doesn't exist). If any code fix was to be offered, I would change assert to something like `Assert.False(item is IInterface)`.

src/xunit.analyzers/Descriptors.cs
jamesqo: Sure. BTW, your project is the first I've seen that has a separate class for descriptors, rather than putting the descriptors in the analyzers. Have you considered the latter option?

marcind: Yes, but you still need to coordinate the analyzer ID numbering. Without a centralized place it would be too easy to have multiple rules use the same id, potentially confusing fixers, ignore pragmas, etc. Plus at the moment this file servers as a listing of all the available rules until I get around to creating some real documentation. But even then this central list would likely remain.

src/xunit.analyzers/AssertIsTypeShouldNotBeUsedForAbstractTypeFixer.cs
marcind: Way I see it, there are 2 ways a user might end up with the analyzer firing for a `IsNotType`:
1. They had just written the assert and we know 100% that it will always succeed, thus it's unnecessary (although could serve documentation purposes, I suppose?)
2. They are refactoring an existing codebase with existing tests and modifying the class hierarchy. As a result a number of existing valid asserts become effective no-ops. Having a fixer that can be applied in batch mode would allow them get the repo back to no warnings. Of course the appropriate fix might be to change the type that's asserted, but having the option might be nice.

marcind: But the codefix you propose also works. It's possible to have multiple fixes for a single issue.

src/xunit.analyzers/Descriptors.cs
jamesqo: @marcind 

> Without a centralized place it would be too easy to have multiple rules use the same id, potentially confusing fixers, ignore pragmas, etc.

Other projects (e.g. [StyleCop](https://github.com/DotNetAnalyzers/StyleCopAnalyzers)) prefix the name of each analyzer with the ID, ensuring that doesn't happen. That also lends itself to being easier to find the file of the analyzer associated with a warning-- people can just go to GitHub, start the fuzzy finder, and type in the ID rather than guessing at the name of the file.

src/xunit.analyzers/AssertIsTypeShouldNotBeUsedForAbstractTypeFixer.cs
jamesqo: @marcind I am hesitant for `Assert.False(item is IInterface)` because I don't know that `is` has exactly the same semantics as `IsAssignableFrom`, and an `IsNotAssignableFrom` might come in the future-- that's why I said *if* any code fix was offered. However, your second point about refactoring a concrete class into an abstract one/an interface is valid, so you've convinced me to create a codefix that removes the assertion.

jamesqo: I have changed my mind on that last sentence-- it's easy for the user to just press Ctrl+X if he needs to delete the assert. I don't feel it's necessary to write a code fix provider for that in this PR.

jamesqo:
@marcind, I've finished responding to your feedback and applied some additional touchups of my own. Feel free to merge if you think this is ready.

marcind (approved):

marcind:
Looks great

PR #67, opened 2017-07-30
Run the built analyzers on the test project

jamesqo:
I ran into an issue during my earlier PR where I forgot to make the test class `public`. This stumped me for a while, so when I realized what the issue was I was surprised because I assumed they were already enabled for the tests here.

This PR causes the freshly-built analyzers to be used against their own test project, so it's easier to find/fix bugs since we'll actually be using them.

### DIFF
--- a/test/xunit.analyzers.tests/TestClassMustBePublicTests.cs
+++ b/test/xunit.analyzers.tests/TestClassMustBePublicTests.cs
@@ -5,7 +5,7 @@ namespace Xunit.Analyzers
 {
     public class TestClassMustBePublicTests
     {
-        private static IEnumerable<object[]> CreateFactsInNonPublicClassCases()
+        public static IEnumerable<object[]> CreateFactsInNonPublicClassCases()
         {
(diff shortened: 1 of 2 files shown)


marcind:
I'm on the fence about this one. One the one hand it would be helpful to find issues in the tests, on the other I'm worried about whether VS keeps that binary in memory and wether rebuilding it all the time would cause problems.

jamesqo:
@marcind 

> on the other I'm worried about whether VS keeps that binary in memory and whether rebuilding it all the time would cause problems.

It doesn't seem to make a difference. I tried opening the solution and rebuilding `xunit.analyzers.tests` 10 times. `devenv` memory usage does not go up. While `VBCSCompiler` memory indeed climbs to ~700MB, it does so regardless of whether I enable the analyzer or not.

jamesqo:
@marcind What about using the latest build of the analyzers from the MyGet feed here: https://myget.org/feed/xunit/package/nuget/xunit.analyzers

bradwilson:
Sorry for the delay. Decided to fix this up and merge it. Thanks!

PR #68, opened 2017-07-30
Add analyzer to ensure all parameters are used within a Theory method

### LINKED ISSUE #1373: [xunit.analyzers] Consider warning when Theory parameter is unused
Sometimes, people convert from `[Fact]` to `[Theory]` but forget to replace the hard-coded values with variables. I recall commenting on a corefx PR of @hughbe's some time ago where he made this mistake, but I can't find it now (there are too many of them :)).

Would it be a good idea to warn when the parameter of a theory is unused, e.g.

```cs
[Theory]
[InlineData(2)]
public void ZeroIsAdditiveIdentity(int n)
{
    Assert.Equal(2, AddZero(2));
}
```

? /cc @marcind

jamesqo:
This fixes https://github.com/xunit/xunit/issues/1373 by adding an analyzer to warn on unused Theory parameters, along with tests for the new analyzer. It uses [`SemanticModel.AnalyzeDataFlow`](https://joshvarty.wordpress.com/2015/02/05/learn-roslyn-now-part-8-data-flow-analysis/) on the theory method and checks if parameters are in `ReadInside` to determine if they are used.

/cc @hughbe, @marcind

### DIFF
--- a/test/xunit.analyzers.tests/TheoryMethodMustUseAllParametersTests.cs
+++ b/test/xunit.analyzers.tests/TheoryMethodMustUseAllParametersTests.cs
@@ -1,0 +1,123 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Xunit.Analyzers
+{
+    public class TheoryMethodMustUseAllParametersTests
+    {
+        private static DiagnosticAnalyzer Analyzer { get; } = new TheoryMethodMustUseAllParameters();
+
+        private static void CheckDiagnostics(IEnumerable<Diagnostic> diagnostics, params (string method, string type, string parameter)[] messageArgs)
+        {
+            var diagnosticArray = diagnostics.ToArray();
+            Assert.Equal(messageArgs.Length, diagnosticArray.Length);
+
+            for (int i = 0; i < messageArgs.Length; i++)
+            {
+                var (method, type, parameter) = messageArgs[i];
+                string message = $"Theory method '{method}' on test class '{type}' does not use parameter '{parameter}'.";
+
+                var diagnostic = diagnosticArray[i];
+                Assert.Equal(message, diagnostic.GetMessage());
+                Assert.Equal("xUnit1026", diagnostic.Id);
+                Assert.Equal(DiagnosticSeverity.Warning, diagnostic.Severity);
+            }
+        }
+
+        [Fact]
+        public async void FindsError_ParameterNotReferenced()
+        {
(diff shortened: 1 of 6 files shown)


test/xunit.analyzers.tests/TheoryMethodMustUseAllParametersTests.cs
```diff
@@ -0,0 +1,140 @@
...
+    [Theory]
+    void TestMethod(int unused)
+    {
+        Console.WriteLine(unused);
+    }
+}");
+        }
+
+        [Fact]
+        public async void DoesNotFindError_ParameterAddressTaken()
```
marcind: This seems like a pretty esoteric case. I guess I'm fine keeping it, thought I'd also like to see one covering a method being invoked on the parameter.

marcind (changes requested):
Thanks so much for the submission. Looks good, just a small nit. Also, it would probably be worth it to add this case to https://github.com/xunit/xunit.analyzers/blob/master/src/xunit.analyzers/FixProviders/RemoveMethodParameterFix.cs.

marcind:
Oh and please squash into one commit.

test/xunit.analyzers.tests/TheoryMethodMustUseAllParametersTests.cs
jamesqo: I don't need this test actually, I only added it because I assumed `DFA.ReadInside` wouldn't include `unused` here and I would have to union it with `DFA.UnsafeAddressTaken` in the analyzer. Turns out I was wrong.

jamesqo:
@marcind 

> Also, it would probably be worth it to add this case to https://github.com/xunit/xunit.analyzers/blob/master/src/xunit.analyzers/FixProviders/RemoveMethodParameterFix.cs.

Sounds like a good idea, I'll do that. Speaking of code fix providers, I noticed that there are no tests in your repo for them, only analyzers. Is there a reason for that?

marcind:
I'm lazy. And I did not see a practical example of how to write them.

test/xunit.analyzers.tests/TheoryMethodMustUseAllParametersTests.cs
jamesqo: @marcind 

> I guess I'm fine keeping it, thought I'd also like to see one covering a method being invoked on the parameter.

There is such a test right above this one.

jamesqo:
@marcind I made the changes you suggested and squashed. BTW, GitHub provides a way to automatically squash all commits when merging a PR: https://github.com/blog/2141-squash-your-commits

marcind:
Unfortunately the xunit org does not have that option enabled at the moment.

PR #69, opened 2017-07-31
Add collection contains analyzer

ErikSchierboom:
Refs: https://github.com/xunit/xunit/issues/1374


jamesqo:
What if someone is making a custom collection and is unit testing their `ICollection<T>.Contains` implementation? Should we care about that scenario?

marcind:
There's a number of ways you can disable an analyzer rule if you're testing something like that. Without specific examples of where this causes a problem I wouldn't worry too much about it.

jamesqo:
@marcind I asked because your analyzers might be used in corefx (the home for all of the framework libraries): https://github.com/dotnet/corefx/issues/22210. Over there they have a lot of collection types and such unit tests for `Contains`.

ErikSchierboom:
@jamesqo As @marcind suggested, I think disabling the analyzer rule would be the best option.

PR #70, opened 2017-08-05
Refactoring: Replace static strings with const strings

jamesqo:
Additionally ran the VS code formatter on all the files in the solution using [this answer](https://stackoverflow.com/a/15458340/4077294). To ignore the whitespace changes, add `&w=1` to the GitHub URL: https://github.com/xunit/xunit.analyzers/pull/70/files?w=1


marcind (changes requested):
I like removing the unsafe compilation support,
I don't mind the whitespace cleanup,
Why change statics to consts?

jamesqo:
@marcind 

> Why change statics to consts?

Because the fields should not be mutable (they are never assigned to). They don't have to be const, I can also change them to `static string ___ { get; }` or `static readonly` if you prefer that-- I've no preference as long as they aren't writeable.

ErikSchierboom:
> Because the fields should not be mutable (they are never assigned to). 

I think this is spot on. If the field should not be changed, why not have it be a `const` value? It better signals intent than `static readonly` IMHO.

marcind:
Closing this issue
- The line endings change was bad, for some reason it was normalizing the affected lines in a way that was resulting in LFs on my machine (I do have autocrlf set to true). Not sure why that was, but I did notice there was no gitattributes file in the repo. That's been [fixed](https://github.com/xunit/xunit.analyzers/commit/dc85ff503644d976043166e51c1158f41ea9e4f9) now, and I redid the [normalization](https://github.com/xunit/xunit.analyzers/commit/4297f95a2b8ab54bfe98547355916ee3307567a5) and [reformatting](https://github.com/xunit/xunit.analyzers/commit/dbc2a61ba8b88215f0519d745e99545d51faee55).
- IMHO there's no practical difference between const and static in this project.
- I cherry-picked the [AllowUnsafe removal change](https://github.com/xunit/xunit.analyzers/commit/942e08ed7e422ef79130d12fc04332ae08785653) a while ago

jamesqo:
@marcind 

> IMHO there's no practical difference between const and static in this project.

What about making the strings `static readonly` instead of `static`, would that be OK?

PR #71, opened 2017-08-05
Add ability to view code coverage

jamesqo:
This PR adds the ability for users to generate and view code coverage reports for `xunit.analyzers`, using OpenCover and ReportGenerator. It's as easy as navigating to the repo root in a command prompt and typing `view-coverage`. The script will call `dotnet test` with OpenCover generating `results.xml`, then generate a report in `<repo root>/reports/xunit.analyzers.tests/` and open it in the browser.

For an infrastructure PR, this contains a lot of code changes. Let me explain why:

- OpenCover mysteriously says everything is 0% coverage when `xunit.analyzers.tests` targets `net452`. So I had to change it to target `netcoreapp1.1` in order for the coverage to work.
  - This necessitated upgrading the version of `Microsoft.CodeAnalysis.CSharp` `xunit.analyzers.tests` used, so I upgraded it to the latest version both there and in `xunit.analyzers`.
    - This required changing `xunit.analyzers` to target `netstandard1.3`, in turn requiring `xunit.analyzers.vsix` to target `net46`.
  - Upgrading to netcoreapp1.1 also made some things not compile (e.g. needed extra `GetTypeInfo()` calls). It also screwed with the references in `CodeAnalyzerHelper`, causing all the tests to fail since `typeof(object).Assembly.Location` is no longer mscorlib.dll but System.Private.CoreLib.ni.dll. I had to rewrite most of the logic there.

### DIFF
--- a/src/xunit.analyzers.vsix/xunit.analyzers.vsix.csproj
+++ b/src/xunit.analyzers.vsix/xunit.analyzers.vsix.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <MinimumVisualStudioVersion>15.0</MinimumVisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+    <TargetFrameworkProfile />
   </PropertyGroup>
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
@@ -15,7 +16,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>xunit.analyzers.vsix</RootNamespace>
     <AssemblyName>xunit.analyzers.vsix</AssemblyName>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
     <GeneratePkgDefFile>false</GeneratePkgDefFile>
     <IncludeAssemblyInVSIXContainer>false</IncludeAssemblyInVSIXContainer>
     <IncludeDebugSymbolsInVSIXContainer>false</IncludeDebugSymbolsInVSIXContainer>
--- a/test/xunit.analyzers.tests/TheoryMethodMustUseAllParametersTests.cs
+++ b/test/xunit.analyzers.tests/TheoryMethodMustUseAllParametersTests.cs
@@ -93,7 +93,7 @@ class TestClass
     [Theory]
     void TestMethod(int foo, int bar, int baz)
     {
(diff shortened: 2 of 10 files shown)


src/xunit.analyzers.vsix/xunit.analyzers.vsix.csproj
```diff
@@ -1,70 +1,71 @@
...
-  </ItemGroup>
-  <ItemGroup>
-    <ProjectReference Include="..\xunit.analyzers\xunit.analyzers.csproj">
-      <Project>{F513916A-00AD-499A-9B76-E52D9FBFCED5}</Project>
-      <Name>xunit.analyzers</Name>
-    </ProjectReference>
-  </ItemGroup>
-  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" Condition="'$(VSToolsPath)' != ''" />
+﻿<?xml version="1.0" encoding="utf-8"?>
```
jamesqo, line 1: https://github.com/xunit/xunit.analyzers/pull/71/files?w=1 to ignore the ws changes

test/xunit.analyzers.tests/TheoryMethodMustUseAllParametersTests.cs
```diff
@@ -93,7 +93,7 @@ class TestClass
     [Theory]
     void TestMethod(int foo, int bar, int baz)
     {
-        Console.WriteLine(bar);
+        TestMethod(bar, bar, bar);
```
jamesqo, line 96: This was necessary since trying to reference `System.Console` in `CodeAnalyzerHelper` created compilation problems. The `Console` type appears to be defined in both `System.Private.CoreLib` and `System.Console`.

jamesqo:
@bradwilson I've changed the test project to target netcoreapp1.1 instead of net452. This has caused CI failures because currently we use the console runner from `tools/net452`. I see there is an `xunit.console.dll` in `tools/netcoreapp1.0`, which I think is the solution to my problem, but there is no EXE file. Do you know how I can use it?

jamesqo:
@marcind You haven't said anything on this one. Would you be willing to merge this if I can resolve conflicts/get CI passing?

marcind:
Sorry, just have not had the time to look at everything. I'd like to try out what impact (if any) this change has on the development workflow. There's also a few things that concern me:
- targeting a some other target than 452. Probably not a big deal but I haven't thought through that yet
- upgrading some of the Roslyn dependencies. Not sure what impact that would have on The analyzers being able to run in VS 2015.
- there's a lot of change so just need to find the time to scrutinize them

bradwilson:
@jamesqo That DLL is there for advanced users (you must know how to correctly use the undocumented `dotnet exec`). Instead, .NET Core users who want functionality like `xunit.console` should be using `dotnet xunit` as illustrated here: https://xunit.github.io/docs/getting-started-dotnet-core.html

bradwilson:
Apologies that this sat here for so long. I've picked up the project from Marcin, and just enabled better PR & CI builds via GitHub actions.

With the new .NET Core-based build system, is this PR still relevant? What is required to enable it in the new build system?

bradwilson:
Closing for age.

PR #72, opened 2017-08-12
Refactoring: Split Extensions into two and move it to Utilities

jamesqo:
The methods in the `Extensions` class seem really random. Apart from them being related to types from Roslyn, they don't really have a coherent theme-- some methods are related to syntax, others to symbols. I split `Extensions` into two new classes, `SyntaxExtensions` and `SymbolExtensions`, and put them in the `Utilities` subdirectory.

/cc @marcind, @ErikSchierboom

### DIFF
--- a/src/xunit.analyzers/SymbolExtensions.cs
+++ b/src/xunit.analyzers/SymbolExtensions.cs
@@ -1,0 +1,47 @@
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+
+namespace Xunit.Analyzers
+{
+    internal static class SymbolExtensions
+    {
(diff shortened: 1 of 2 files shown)


marcind (changes requested):
I typically start a project simple until the right organizational patterns emerge. Looks like its time for those extension methods. However, I would prefer the namespace not to change so that it's easier for intellisense to kick in. Otherwise looks good.

jamesqo:
> However, I would prefer the namespace not to change so that it's easier for intellisense to kick in. Otherwise looks good.

I disagree a bit there, it's a bit weird if some 'utility' classes are in `Utilities/` and some are not, but sure, I've changed the namespace. Feel free to merge.

ErikSchierboom (approved):

marcind (approved):

PR #73, opened 2017-08-13
Refactoring: Cleaner way to require that types be present at start of a compilation

jamesqo:
I noticed that something like

```cs
context.RegisterCompilationStartAction(compilationStartContext =>
{
    var factType = compilationStartContext.Compilation.GetTypeByMetadataName(Constants.Types.XunitFactAttribute);
    if (factType == null)
        return;

    compilationStartContext.RegisterAnotherAction(anotherContext => /* use factType */);
}
```

seemed to be a very common pattern in the codebase. I decided to refactor this since it was very noisy and created an extra level of indentation/callback nesting. The above code now becomes:

```cs
context.RequireTypes(Constants.Types.XunitFactAttribute).RegisterAnotherAction(anotherContext =>
{
    var factType = anotherContext.Compilation.GetFactAttributeType();
    /* use factType */
});
```

/cc @marcind, @ErikSchierboom 

**edit:** `w=1` will help reviewing the changes. https://github.com/xunit/xunit.analyzers/pull/73/files?w=1


src/xunit.analyzers/Utilities/TypeHierarchyComparer.cs
```diff
@@ -12,9 +12,9 @@ internal class TypeHierarchyComparer : IComparer<ITypeSymbol>
 
         public int Compare(ITypeSymbol x, ITypeSymbol y)
         {
-            if (x == null || x.TypeKind != TypeKind.Class)
+            if (x?.TypeKind != TypeKind.Class)
                 throw new ArgumentException("The argument must be a class", nameof(x));
-            if (y == null || x.TypeKind != TypeKind.Class)
```
jamesqo, line 17: Spotted a bug here while I was refactoring: it should be `y.TypeKind != TypeKind.Class`. I fixed it.

jamesqo:
@marcind Just curious, why don't you skip `RegisterCompilationStartAction` entirely and just write

```cs
// context is AnalysisContext
context.RegisterSymbolAction(symbolContext =>
{
    var factType = symbolContext.Compilation.GetTypeByMetadataName(Constants.Types.XunitFactAttribute);
});
```

? Did you run into performance problems from calling `GetTypeByMetadataName` unnecessarily?

marcind:
I'm really sorry I have not been able to respond to this earlier as it looks like you've put a lot of time and effort into this. However, I'm not able to accept this PR as it takes the codebase in a different direction than what I've been thinking of. It really pains me to turn down well-intentioned effort but at the same time I need to be blunt: I would much rather review and accept PRs that contain new features and bug fixes (as taken from the issue tracker) than refactorings and stylistic changes. In the future it would be more productive if you could give me a heads up when you're thinking about making any medium-sized or larger changes and preferably open an issue for discussion first. Once again, sorry.

jamesqo:
@marcind No problem, I understand. I will keep in mind to ask you about such changes beforehand. I do have a couple of bug/feature-oriented PRs coming up, I just thought it would be easier to think about them after refactoring a little. Thanks for giving me a better understanding of your intentions for this project.

PR #76, opened 2017-08-15
Upgrade to Roslyn 1.2.2, enable concurrent execution in all analyzers

jamesqo:
Fixes xunit/xunit#1403.

Changes:

- Upgrade M.CA.CS.W to 1.2.2, M.C.A to 1.2.0-beta2
- Enable concurrent execution in all analyzers
- `SyntaxNodeAnalysisContext.Compilation` is now available, which is shorthand for `context.SemanticModel?.Compilation`.
- Other minor tweaks

/cc @marcind

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 xUnit.net is a free, open source, community-focused unit testing tool for the .NET Framework. Written by the original inventor of NUnit v2, xUnit.net is the latest technology for unit testing C#, F#, VB.NET and other .NET languages. xUnit.net works with ReSharper, CodeRush, TestDriven.NET and Xamarin. It is part of the [.NET Foundation](https://www.dotnetfoundation.org/), and operates under their [code of conduct](https://www.dotnetfoundation.org/code-of-conduct). It is licensed under [Apache 2](https://opensource.org/licenses/Apache-2.0) (an OSI approved license).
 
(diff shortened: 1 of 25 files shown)


ErikSchierboom:
@jamesqo Have you noticed any performance improvement after doing these changes?

jamesqo:
@ErikSchierboom Can you point to any large solutions that I can test this on?

ErikSchierboom:
@jamesqo I'm afraid not, it's my work solution which I cannot share :(

jamesqo:
@sharwell What tools do you typically use to profile your analyzers? Also, is there some way to test out my changes against the Roslyn unit tests?

sharwell:
We wrote StyleCopTester for this. It's a small command line utility in our repository that you could modify to run xunit analyzers on a solution instead.

jamesqo:
I was planning to do the profiling but I guess @marcind doesn't want to wait :smile:

marcind:
It would still be great if you could profile so we can have an
understanding on the scale of the improvement.

sharwell:
Profiling will be a bit misleading in some configurations. It's fine for finding ways to improve your analysis but you'll have trouble drawing conclusions relating to things like the use of concurrent analysis. I'm on a phone now but if you ask me about it on Aug 28 I'll be able to explain much more.

PR #78, opened 2017-08-18
[WIP] Documentation for rules

jamesqo:
This PR begins the process of adding documentation for each rule. The docs are based off the StyleCopAnalyzers documentation here: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/tree/master/documentation, where each rule gets its own markdown file named after its ID. I'd like to get everyone's feedback before proceeding with further documentation, especially on the rule template.

Contributes to https://github.com/xunit/xunit/issues/1262

/cc @marcind, @ErikSchierboom

### DIFF
--- a/docs/xUnit2004.md
+++ b/docs/xUnit2004.md
@@ -1,0 +1,66 @@
+## xUnit2004
+
+<table>
+<tr>
+  <td>Name</td>
+  <td>AssertEqualShouldNotBeUsedForBoolLiteralCheck</td>
+</tr>
+<tr>
+  <td>ID</td>
+  <td>xUnit2004</td>
+</tr>
+<tr>
+  <td>Category</td>
+  <td>Assertions</td>
+</tr>
+<tr>
+  <td>Severity</td>
+  <td>Warning</td>
+</tr>
+</table>
+
+## Cause
+
+A violation of this rule occurs when:
+
+- `Assert.Equal`, `Assert.NotEqual`, `Assert.StrictEqual`, or `Assert.NotStrictEqual` is used
+- The expected value is a `true` or `false` literal
+
+## Reason for rule
+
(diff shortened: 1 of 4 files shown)


docs/xUnit2004.md
```diff
@@ -0,0 +1,66 @@
...
+```csharp
+Assert.True(2 + 2 == 4);
+Assert.False(2 + 2 == 5);
+```
+
+## How to fix violations
+
+### For `Equal` and `StrictEqual`
+
+- `Assert.{Strict}Equal(true, b)` => `Assert.True(b)`
```
ErikSchierboom, line 53: I'm not sure the `{Strict}` bit makes sense here. Sure it is shorter, but I think two separate lines (one for `Assert.Equal` and one for `Assert.StrictEqual`) would be more readable.

docs/xUnit2004.md
```diff
@@ -0,0 +1,66 @@
...
+- `Assert.{Strict}Equal(false, b)` => `Assert.False(b)`
+
+### For `NotEqual` and `NotStrictEqual`
+
+- `Assert.Not{Strict}Equal(true, b)` => `Assert.False(b)`
```
ErikSchierboom, line 58: I'm not sure the `{Strict}` bit makes sense here. Sure it is shorter, but I think two separate lines (one for `Assert.Equal` and one for `Assert.StrictEqual`) would be more readable.

ErikSchierboom (commented):
Overall I think this is fantastic! The format is easy to read and very clear. Great work!

jamesqo:
@marcind Thanks for merging this. In the future, please wait until I un-mark a PR as [WIP] before merging it. I was planning some other changes in addition to responding to PR feedback (which it looks like you took care of yourself), but was waiting for a go-ahead at https://github.com/xunit/xunit/issues/1262. I'll put those changes in a new PR.

PR #80, opened 2017-08-23
Add xUnit1027: TestMethodShouldNotHaveReturnType

### LINKED ISSUE #1415: Issue diagnostic when test method return type is not void or Task
For example, the following code should raise a diagnostic with the squiggly line under `int`:

```cs
[Fact] public int Test() { return 0; }
```

Also, in the case of `Task`, the return type must be exactly `Task` and not a derived type. This should also raise a diagnostic:

```cs
[Fact] public async Task<int> TestAsync() { return 0; }
```

@marcind Do you think this should be a warning or an error?

jamesqo:
Fixes xunit/xunit#1415

### DIFF
--- a/src/xunit.analyzers/TestMethodShouldNotHaveReturnType.cs
+++ b/src/xunit.analyzers/TestMethodShouldNotHaveReturnType.cs
@@ -1,0 +1,40 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Xunit.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class TestMethodShouldNotHaveReturnType : XunitDiagnosticAnalyzer
+    {
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
+            ImmutableArray.Create(Descriptors.X1027_TestMethodShouldNotHaveReturnType);
+
+        internal override void AnalyzeCompilation(CompilationStartAnalysisContext compilationStartContext, XunitContext xunitContext)
+        {
+            var compilation = compilationStartContext.Compilation;
+            var taskType = compilation.GetTypeByMetadataName(Constants.Types.SystemThreadingTasksTask);
+
+            compilationStartContext.RegisterSymbolAction(symbolContext =>
+            {
+                var methodSymbol = (IMethodSymbol)symbolContext.Symbol;
+                // NOTE: This analyzer should fire even if the return type is a derived type of Task, such as Task<int>.
+                if (methodSymbol.ReturnsVoid || (taskType != null && methodSymbol.ReturnType == taskType))
+                    return;
+
+                var attributes = methodSymbol.GetAttributes();
+                if (!attributes.ContainsAttributeType(xunitContext.FactAttributeType, exactMatch: true) &&
+                    !attributes.ContainsAttributeType(xunitContext.TheoryAttributeType, exactMatch: true))
+                    return;
+
+                symbolContext.ReportDiagnostic(Diagnostic.Create(
+                    Descriptors.X1027_TestMethodShouldNotHaveReturnType,
+                    methodSymbol.Locations.First(),
+                    methodSymbol.Name,
+                    methodSymbol.ContainingType.Name));
+            }, SymbolKind.Method);
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


marcind (changes requested):
There are merge conflicts and descriptor name change should be a separate issue.

marcind:
FYI, I already checked in the descriptor name change for  [TheoryMethodShouldUseAllParameters](https://github.com/xunit/xunit.analyzers/commit/3f055d5158351a12ae975c722929884a9d1f41d2)

jamesqo:
@marcind Great, I'll back out those changes here then.

jamesqo:
@marcind I've un-WIPed this PR, so feel free to review when you have time.

src/xunit.analyzers/TestMethodShouldNotHaveReturnType.cs
```diff
@@ -0,0 +1,40 @@
...
+            compilationStartContext.RegisterSymbolAction(symbolContext =>
+            {
+                var methodSymbol = (IMethodSymbol)symbolContext.Symbol;
+                // NOTE: This analyzer should fire even if the return type is a derived type of Task, such as Task<int>.
+                if (methodSymbol.ReturnsVoid || (taskType != null && methodSymbol.ReturnType == taskType))
+                    return;
+
+                var attributes = methodSymbol.GetAttributes();
+                if (!attributes.ContainsAttributeType(xunitContext.FactAttributeType, exactMatch: true) &&
+                    !attributes.ContainsAttributeType(xunitContext.TheoryAttributeType, exactMatch: true))
```
jamesqo, line 29: @marcind I'm not very familiar with the xunit extensibility model. I know that people can extend `Fact` and create their own test method attributes, but I'm unsure whether it would be possible to have a fact where there's a good reason for having a non-void/Task return type. Should I keep the current implementation (only fires for Fact/Theory) or go with

```cs
// Covers Fact, Theory, and other custom attributes derived from Fact
if (!attributes.ContainsAttributeType(xunitContext.FactAttributeType))
```

?

marcind: I guess it's possible someone could be doing something with the result of the test method invocation. But part of me thinks we should just run this on any method decorated with `Fact` or a derived attribute just to see if someone has that scenario. Worst case, the rule can be disabled by the user until we sort it out.

src/xunit.analyzers/TestMethodShouldNotHaveReturnTypeFixer.cs
```diff
@@ -0,0 +1,71 @@
...
+                        ct => ChangeReturnTypeAsync(context.Document, methodDeclaration, taskSyntax, ct),
+                        equivalenceKey: title),
+                    context.Diagnostics);
+            }
+        }
+
+        private async Task<Document> ChangeReturnTypeAsync(Document document, MethodDeclarationSyntax methodDeclaration, TypeSyntax newReturnType, CancellationToken ct)
+        {
+            var editor = await DocumentEditor.CreateAsync(document, ct).ConfigureAwait(false);
+            editor.ReplaceNode(methodDeclaration.ReturnType, newReturnType);
```
marcind, line 61: This only changes the methods return type. Any `return [expression];` statements remain in the test resulting in code that does not compile (instead of just being a warning before). It would be nice if the fixer could convert those to simple `return;` statements (and for bonus points, remove any redundant ones where the control flow in the method would cause an implicit return anyway).

marcind (commented):
The fixer can result in code that does not compile. Wonder if we can improve on that.

src/xunit.analyzers/TestMethodShouldNotHaveReturnTypeFixer.cs
jamesqo: Ooh, this sounds like an interesting challenge. I expect it'll take me some time to implement this, but this sounds like a good idea for the codefix.

bradwilson:
Sorry for the delay in getting back to this.

Upon reviewing the code and the discussion, I have two opinions:

- I want to be more conservative than Marcin did, I think, and am fine with it being explicitly limited to `[Fact]` and `[Theory]`.
- We definitely should be changing return statements in the fixer.

I had a few additional thoughts as well:

- I just merged in another rule as `xUnit1027` so you'll need to move to `xUnit1028`
- The testing framework in use now is significantly different than it was in 2017, so tests will need to be ported (and the fixer moved to the fixer project, since analyzers and fixers live in different assemblies now)
- Once this change is merged, we should open an issue for analysis for v3 to also allow returning `ValueTask` (but definitely limit this to tests that link in v3, and not v2).

I understand if your interest in this is long gone (again, apologies for the delays). If that's the case, feel free to just chime in and say you'd rather not make the changes and I'll look for some time in the future to do them myself, because I feel like this is a valuable feature.

Thanks!

bradwilson:
Closing for age.

PR #81, opened 2017-08-29
Finish documentation work from #78

jamesqo:
This PR finishes up the work from #78 by moving the rule docs into a `reference/` subdir per https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Documenting%20your%20analyzers.md, and adds help link URIs to point to the documentation. It also adds an `adding-new-rules.md` document that walks contributors through the process of adding a new rule.

/cc @marcind, @ErikSchierboom


marcind:
Thanks for submitting this but I have something slightly different in mind for how I would like the documentation to be structured.

jamesqo:
@marcind OK, I'm fine with keeping the docs out of a `reference/` directory. Are you still planning to merge the `helpLinkUri` changes in your own commits? Do you think adding-new-rules.md is worthwhile or is it too bureaucratic for you?

marcind:
https://github.com/xunit/xunit.analyzers/commit/c31c360e2594ce3fc8193cf8ac23c0fdd22b587b includes the help links in the descriptors, the path is slightly different.

Adding-new-rules.md is more of a contributor guide, whereas I see xunit.github.io as end-user documentation. However, there's hardly any contributors, so not sure if it's worth maintaining.

PR #82, opened 2017-09-22
Fixes false positives with public methods which are not allowed

4creators:
as test methods in xunit: "public static" and "public abstract"

PublicMethodShouldBeMarkedAsTest greedily issues warning for 'public static' and 'public abstract'
methods although they can not be used as a test method in xunit framework. Introduced code changes
verify if public method is abstract or is static and if detected skip further checks.

Fixes https://github.com/xunit/xunit/issues/1466

### DIFF
--- a/src/xunit.analyzers.vsix/source.extension.vsixmanifest
+++ b/src/xunit.analyzers.vsix/source.extension.vsixmanifest
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
   <Metadata>
-    <Identity Id="xunit.analyzers..d09124d1-9c5e-40de-b1de-35120196c3e0" Version="1.0.1" Language="en-US" Publisher="Marcin Dobosz"/>
+    <Identity Id="xunit.analyzers..d09124d1-9c5e-40de-b1de-35120196c3e0" Version="1.0" Language="en-US" Publisher="Marcin Dobosz"/>
     <DisplayName>xunit.analyzers</DisplayName>
     <Description xml:space="preserve">Code Analyzers for projects using xUnit.net that help find and fix frequent issues when writing tests.</Description>
   </Metadata>
--- a/src/xunit.analyzers.vsix/source.extension.vsixmanifest
+++ b/src/xunit.analyzers.vsix/source.extension.vsixmanifest
@@ -1,21 +1,21 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
-    <Metadata>
-        <Identity Id="xunit.analyzers..d09124d1-9c5e-40de-b1de-35120196c3e0" Version="1.0.1" Language="en-US" Publisher="Marcin Dobosz"/>
-        <DisplayName>xunit.analyzers</DisplayName>
-        <Description xml:space="preserve">Code Analyzers for projects using xUnit.net that help find and fix frequent issues when writing tests.</Description>
-    </Metadata>
-    <Installation>
-        <InstallationTarget Version="[14.0,]" Id="Microsoft.VisualStudio.Community" />
-    </Installation>
-    <Dependencies>
-        <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" d:Source="Manual" Version="[4.5,)" />
-    </Dependencies>
-    <Assets>
-        <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="xunit.analyzers" Path="|xunit.analyzers|"/>
-        <Asset Type="Microsoft.VisualStudio.Analyzer" d:Source="Project" d:ProjectName="xunit.analyzers" Path="|xunit.analyzers|"/>
-    </Assets>
-    <Prerequisites>
-        <Prerequisite Id="Microsoft.VisualStudio.Component.CoreEditor" Version="[15.0.26208.0,16.0)" DisplayName="Visual Studio core editor" />
-    </Prerequisites>
+  <Metadata>
+    <Identity Id="xunit.analyzers..d09124d1-9c5e-40de-b1de-35120196c3e0" Version="1.0.1" Language="en-US" Publisher="Marcin Dobosz"/>
+    <DisplayName>xunit.analyzers</DisplayName>
+    <Description xml:space="preserve">Code Analyzers for projects using xUnit.net that help find and fix frequent issues when writing tests.</Description>
+  </Metadata>
+  <Installation>
+    <InstallationTarget Version="[14.0,]" Id="Microsoft.VisualStudio.Community" />
+  </Installation>
+  <Dependencies>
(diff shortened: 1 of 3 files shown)


4creators:
PTAL @marcind

src/xunit.analyzers.vsix/source.extension.vsixmanifest
```diff
@@ -1,21 +1,21 @@
...
+        <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" d:Source="Manual" Version="[4.5,)" />
+    </Dependencies>
+    <Assets>
+        <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="xunit.analyzers" Path="|xunit.analyzers|"/>
+        <Asset Type="Microsoft.VisualStudio.Analyzer" d:Source="Project" d:ProjectName="xunit.analyzers" Path="|xunit.analyzers|"/>
+    </Assets>
+    <Prerequisites>
+        <Prerequisite Id="Microsoft.VisualStudio.Component.CoreEditor" Version="[15.0.26208.0,16.0)" DisplayName="Visual Studio core editor" />
+    </Prerequisites>
 </PackageManifest>
```
bradwilson: Please do not reformat files. The entire file looks changed because of spacing differences.

src/xunit.analyzers/PublicMethodShouldBeMarkedAsTest.cs
```diff
@@ -42,7 +43,7 @@ internal override void AnalyzeCompilation(CompilationStartAnalysisContext compil
                     symbolContext.CancellationToken.ThrowIfCancellationRequested();
 
                     var method = (IMethodSymbol)member;
-                    if (method.MethodKind != MethodKind.Ordinary)
+                    if (method.MethodKind != MethodKind.Ordinary || method.IsStatic || method.IsAbstract)
```
bradwilson: xUnit.net does support public static test methods.

bradwilson (changes requested):

src/xunit.analyzers/PublicMethodShouldBeMarkedAsTest.cs
4creators: Yeah, I have tested initially and xunit failed to detect that but using xunit.analyzers.tests I got it as You said. My wrong

src/xunit.analyzers.vsix/source.extension.vsixmanifest
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
   <Metadata>
-    <Identity Id="xunit.analyzers..d09124d1-9c5e-40de-b1de-35120196c3e0" Version="1.0" Language="en-US" Publisher="Marcin Dobosz"/>
+    <Identity Id="xunit.analyzers..d09124d1-9c5e-40de-b1de-35120196c3e0" Version="1.0.1" Language="en-US" Publisher="Marcin Dobosz"/>
```
marcind: Why is this file changed at all? Please undo.

4creators: Thought that xunit has semantic versioning. Change in error/warning detection syntax changes behaviour so I have incremented build version change.

Obviously will change it as requested.

4creators:
Addressed all review remarks 
@bradwilson @marcind

src/xunit.analyzers/PublicMethodShouldBeMarkedAsTest.cs
```diff
@@ -42,7 +43,7 @@ internal override void AnalyzeCompilation(CompilationStartAnalysisContext compil
                     symbolContext.CancellationToken.ThrowIfCancellationRequested();
 
                     var method = (IMethodSymbol)member;
-                    if (method.MethodKind != MethodKind.Ordinary)
+                    if (method.MethodKind != MethodKind.Ordinary || method.IsAbstract)
```
jamesqo: Is this ever true? We returned earlier if `type.IsAbstract` was true, so logically this must be a concrete type and none of its members can be abstract.

jamesqo: Maybe the best fix would be to remove the `type.IsAbstract` check above and keep this as-is.

test/xunit.analyzers.tests/PublicMethodShouldBeMarkedAsTestTests.cs
```diff
@@ -38,6 +39,18 @@ public class PublicMethodShouldBeMarkedAsTestTests
         }
 
         [Fact]
+        public async void DoesNotFindErrorForPublicAbstractMethod()
+        {
+            var diagnostics = await CodeAnalyzerHelper.GetDiagnosticsAsync(analyzer,
+@"public abstract class TestClass {
+    [Xunit.Fact] public void TestMethod() { }
+    public abstract void AbstractTestError();
```
jamesqo: Nit: Rename to `AbstractMethod`, since it isn't a test method and it doesn't raise an error.

test/xunit.analyzers.tests/PublicMethodShouldBeMarkedAsTestTests.cs
```diff
@@ -38,6 +39,18 @@ public class PublicMethodShouldBeMarkedAsTestTests
...
+}");
+
+            Assert.Empty(diagnostics);
+        }
+
```
jamesqo, line 52: Would be good to also add this test:

```cs
        [Fact]
        public async void DoesNotFindErrorForPublicAbstractMethodMarkedWithFact()
        {
            var diagnostics = await CodeAnalyzerHelper.GetDiagnosticsAsync(analyzer,
@"public abstract class TestClass {
    [Xunit.Fact] public void TestMethod() { }
    [Xunit.Fact] public abstract void AbstractTestMethod();
}");
            Assert.Empty(diagnostics);
        }
```

I can't test right now but that should work fine at runtime, since the `[Fact]` attribute is inherited on overriding members.

src/xunit.analyzers/PublicMethodShouldBeMarkedAsTest.cs
4creators: Sorry for late replay, I was stuck with other work for a while.
This way it is set it can skip diagnostics in partially invalid code during code editing i.e.  concrete class with abstract method defined or skip analysis in abstract class for public concrete methods.

test/xunit.analyzers.tests/PublicMethodShouldBeMarkedAsTestTests.cs
4creators: Will do

test/xunit.analyzers.tests/PublicMethodShouldBeMarkedAsTestTests.cs
4creators: OK it makes sense

4creators:
PTAL
@bradwilson @marcind @jamesqo 
I left one unresolved problem to decide see: https://github.com/xunit/xunit.analyzers/pull/82#discussion_r144719822

src/xunit.analyzers/PublicMethodShouldBeMarkedAsTest.cs
jamesqo: OK, it's fine to leave this as-is then, but can you add a comment about it?

jamesqo:
After you leave a comment LGTM

bradwilson (approved):
👍 once the requested comment has been made

4creators:
OK All feedback addressed 😄 
PTAL @bradwilson

marcind:
Thanks for the submission

PR #87, opened 2017-11-13
Update TestCaseMustBeLongLivedMarshalByRefObjectFixerTests.cs

IEvangelist:


bradwilson:
Uhhh, no. 😛

IEvangelist:
What?! C'mon now... this is a valid improvement. https://msdn.microsoft.com/en-us/magazine/dn818493.aspx, it is still an `async void`.

bradwilson:
We support `async void`; in fact, I'll wager there are more `async void` unit tests in this code than `async Task`. There's zero reason to convert them.

IEvangelist:
Just now saw that... interesting, thanks for enlightening me. :+1:

PR #88, opened 2017-11-19
#1561: xUnit1026 should not be raised for parameters named with a starting _

michalfi:


jamesqo:
@marcind I'm not sure if we should merge this. https://github.com/xunit/xunit/issues/1561#issuecomment-345537431

marcind:
Sorry for taking a while to get here, but due to the reasons explained in https://github.com/xunit/xunit/issues/1561#issuecomment-372961831 I decided not to take this PR.

PR #89, opened 2017-11-20
WIP: Make it easier for analyzers to target specific references

bradwilson:
This is a minor UX overhaul on `XunitDiagnosticAnalyzer`:

- Move the core reference verification into `ShouldAnalyze`
- Make it easy for analyzers to know which references are present
- Separate the big type list into context objects (right now, one each for `xunit.abstractions`, `xunit.core`, and `xunit.execution.*`)

This is mostly just a SWAG at the changes. If this is a desired direction, then there are a couple more things I'd want to do:

- Move everything from `XunitCapabilities` into `CoreContext` (since they are all capabilities that depend on a reference to `xunit.core` anyway)
- Either ypdate the context properties to also use `Lazy<T>`, or move to `GetXyz` methods to make it clear that they're expensive


src/xunit.analyzers/Analysis/XunitContext.cs
```diff
@@ -0,0 +1,40 @@
...
+        readonly Lazy<bool> hasAbstractionsReference;
+        readonly Lazy<bool> hasCoreReference;
+        readonly Lazy<bool> hasExecutionReference;
+
+        internal XunitContext(Compilation compilation, XunitCapabilitiesFactory capabilitiesFactory)
+        {
+            Capabilities = capabilitiesFactory(compilation);
+            Compilation = compilation;
+
+            hasAbstractionsReference = new Lazy<bool>(() => Compilation.ReferencedAssemblyNames.FirstOrDefault(x => x.Name == "xunit.abstractions") != null);
```
jamesqo: nit: Why not just `Any(x => x.Name == "xunit.abstractions")`?

marcind (approved):
I like the direction. Full steam ahead.

src/xunit.analyzers/Analysis/XunitContext.cs
bradwilson: Because I'm an idiot and didn't think of it. 😝

marcind:
BTW, I prefer the `Lazy<T>` approach to storing per-compilation references to metadata.

PR #92, opened 2018-01-07
Ignore 1013

adamralph:
Fixes https://github.com/xunit/xunit/issues/1611

### DIFF
--- a/src/xunit.analyzers/PublicMethodShouldBeMarkedAsTest.cs
+++ b/src/xunit.analyzers/PublicMethodShouldBeMarkedAsTest.cs
@@ -50,10 +50,13 @@ namespace Xunit.Analyzers
                     if (method.MethodKind != MethodKind.Ordinary || method.IsAbstract)
                         continue;
 
-                    var isTestMethod = method.GetAttributes().ContainsAttributeType(xunitContext.Core.FactAttributeType);
+                    var attributes = method.GetAttributes();
+                    var isTestMethod = attributes.ContainsAttributeType(xunitContext.Core.FactAttributeType);
                     hasTestMethods = hasTestMethods || isTestMethod;
 
-                    if (isTestMethod)
+                    if (isTestMethod ||
+                        attributes.Any(attribute => attribute.AttributeClass.GetAttributes()
+                            .Any(att => att.AttributeClass.Name.EndsWith("IgnoreXunitAnalyzersRule1013Attribute"))))
                         continue;
 
                     if (method.DeclaredAccessibility == Accessibility.Public &&
--- a/test/xunit.analyzers.tests/PublicMethodShouldBeMarkedAsTestTests.cs
+++ b/test/xunit.analyzers.tests/PublicMethodShouldBeMarkedAsTestTests.cs
@@ -127,6 +127,20 @@ public class TestClass : IntermediateClass {
             Assert.Empty(diagnostics);
         }
 
+        [Fact]
+        public async void DoesNotFindErrorForPublicMethodMarkedWithAttributeWhichIsMarkedWithIgnoreXunitAnalyzersRule1013()
+        {
+            var diagnostics = await CodeAnalyzerHelper.GetDiagnosticsAsync(analyzer,
+@"public class IgnoreXunitAnalyzersRule1013Attribute : System.Attribute { }
+
+[IgnoreXunitAnalyzersRule1013]
+public class CustomTestTypeAttribute : System.Attribute { }
+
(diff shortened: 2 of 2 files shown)


clairernovotny:
I like the idea of enabling custom attributes to suppress certain analyzers. My concern here is that this is too hard-coded as it's only designed to suppress a specific rule. I would prefer to see a mechanism that was trivially extensible for a custom attribute to suppress other analyzer rules too.

jamesqo:
I think it would be better to have an attribute on the attribute:

```cs
[DisableXunitAnalyzers(1013, 1014, ...)]
public sealed class CustomTestTypeAttribute : Attribute {}
```

bradwilson:
I also like an attribute better, because it can be applied more places than an interface.

jamesqo:
There is one issue: with the current name, `DisableXunitAnalyzers`, one could picture people trying to apply it directly to a class/method:

```cs
[DisableXunitAnalyzers(1013)]
public class TestClass
{
    // or
    [DisableXunitAnalyzers(1013)]
    public void TestMethod() { }
}
```

when they should be doing `#pragma disable warning` in such cases instead. This new attribute is only intended to be applied to other attributes that are applied to test classes/methods, not directly applied to those classes/methods.

Should we:

- Allow this too

- Have an analyzer to warn against this

- Rename to `AttributeDisablesXunitAnalyzers(...)` to prevent ambiguity

?

bradwilson:
Other than redundancy w/ other disabling mechanisms, I'm not sure I find it objectionable if people were to also use it in that way. They're not likely to know about it (since I would assume it would live in the `Xunit.Sdk` namespace, which only extenders would use).

bradwilson:
The difficulty is going to know how to look for it, because it's not necessarily intended to be attached to the thing that's being warned about (the method) but rather an adornment (the attribute), which means you probably can't have a general purpose implementation.

adamralph:
I've rebased, and switched from using an interface to using an attribute. Note that I am still duck-typing the attribute, so there's no need for it to be defined in xunit itself.

src/xunit.analyzers/PublicMethodShouldBeMarkedAsTest.cs
```diff
@@ -50,10 +50,13 @@ internal override void AnalyzeCompilation(CompilationStartAnalysisContext compil
...
                         continue;
 
-                    var isTestMethod = method.GetAttributes().ContainsAttributeType(xunitContext.Core.FactAttributeType);
+                    var attributes = method.GetAttributes();
+                    var isTestMethod = attributes.ContainsAttributeType(xunitContext.Core.FactAttributeType);
                     hasTestMethods = hasTestMethods || isTestMethod;
 
-                    if (isTestMethod)
+                    if (isTestMethod ||
+                        attributes.Any(attribute => attribute.AttributeClass.GetAttributes()
```
marcind, line 58: Will this still work with attribute class hierarchies where only the base type is annotated with the magical attribute but some derived attribute type is not (but inherits the annotation)? Also, what happens if the attribute is marked to not be inherited via [AttributeUsageAttribute.Inherited](https://msdn.microsoft.com/en-us/library/system.attributeusageattribute.inherited(v=vs.110).aspx)

marcind (commented):
Sorry for not looking at this sooner. Some questions about attribute inheritance etc. Let's try to make a push to get this into the next build of analyzers so they can be included with the xunit 2.4 betas.

src/xunit.analyzers/PublicMethodShouldBeMarkedAsTest.cs
adamralph: > Will this still work with attribute class hierarchies where only the base type is annotated with the magical attribute but some derived attribute type is not (but inherits the annotation)?

No, but I don't think this worth worrying about. It's trivial to annotate the derived class if required, and I think this somewhat of an edge case anyway.

adamralph:
> Sorry for not looking at this sooner.

👍 no stress.

> Let's try to make a push to get this into the next build of analyzers so they can be included with the xunit 2.4 betas.

🎉 that would be awesome.

src/xunit.analyzers/PublicMethodShouldBeMarkedAsTest.cs
marcind: Would you mind adding tests that document this design choice?

adamralph: @marcind I added another commit which adds such a test (I also rebased the whole thing).

adamralph:
@marcind this should be ready to go now. It looks like the build failed because of a connectivity problem when downloading the NuGet CLI.

marcind (approved):

PR #94, opened 2018-02-03
Docs: xUnit2016 rule.

wachulski:


docs/_rules/xUnit2016.md
```diff
@@ -5,37 +5,35 @@ category: Assertions
...
 ## Cause
 
-A concise-as-possible description of when this rule is violated. If there's a lot to explain, begin with "A violation of this rule occurs when..."
+Asserting on equality of two double or decimal values was declared with precision out of the acceptable range.
 
 ## Reason for rule
 
-Explain why the user should care about the violation.
+`Assert.Equals` for doubles and decimals allows to declare the precision of the check which must be within an 
+arbitrary range: [0..15] for doubles and [0..28] for decimals. This is due to the fact that `Assert.Equals`
```
marcind: The allowed range is not arbitrary or random but rather dictated by how floating point numbers work. This can be simplified to something like

`Assert.Equals` uses [`System.Math.Round`](https://docs.microsoft.com/en-us/dotnet/api/system.math.round?view=netframework-4.7.1) internally which imposes limits on the precision parameter of [0..15] for doubles and [0..28] for decimals.

marcind (changes requested):
Small nit about 'arbitrary'

docs/_rules/xUnit2016.md
wachulski: Changed

marcind (approved):

PR #95, opened 2018-02-06
Fix xUnit1000 for partial classes.

### LINKED ISSUE #1602: xUnit1000: Analyzer does not recognize `partial class` as public if `public` modifier is not repeated
Given the following snippet, the xUnit analyzer will erroneously emit an error *xUnit1000: Test classes must be public*. This is obviously wrong, since the test class in question *is* public:

```csharp
public partial class Tests
{
}

partial class Tests
//            ^^^^^ Error xUnit1000: Test classes must be public
{
	[Fact]
	public void Test()
	{
	}
}
```

xunit-analyzer-partial-class

(I would normally have the two `Tests` class declarations in separate files, but for brevity's sake I have put them together here.)

I am using packages `xunit` 2.3.1 and `xunit.analyzers` 0.7.0 (though the error appears to be present in 0.8.0, too).

wachulski:
TestClassMustBePublic analyzer has been rewritten to analyze class
declaration symbols rather than syntaxes.

The fixer remains unchanged as it corrects code to have proper semantic.

Closes xunit/xunit#1602

### DIFF
--- a/test/xunit.analyzers.tests/TestClassMustBePublicTests.cs
+++ b/test/xunit.analyzers.tests/TestClassMustBePublicTests.cs
@@ -1,25 +1,146 @@
-﻿using Microsoft.CodeAnalysis.Diagnostics;
+﻿using System.Collections.Generic;
+using Microsoft.CodeAnalysis.Diagnostics;
 
 namespace Xunit.Analyzers
 {
     public class TestClassMustBePublicTests
     {
-        readonly DiagnosticAnalyzer analyzer = new TestClassMustBePublic();
+        private readonly DiagnosticAnalyzer analyzer = new TestClassMustBePublic();
+
+        private static IEnumerable<object[]> CreateFactsInNonPublicClassCases()
+        {
+            foreach (var factAttribute in new[] {"Xunit.Fact", "Xunit.Theory"})
+            {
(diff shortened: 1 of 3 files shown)


test/xunit.analyzers.tests/TestClassMustBePublicTests.cs
```diff
@@ -1,25 +1,142 @@
...
+            var diagnostics = await CodeAnalyzerHelper.GetDiagnosticsAsync(analyzer, source);
+
+            Assert.Empty(diagnostics);
+        }
+
+        [Theory]
+        [MemberData(nameof(CreateFactsInNonPublicClassCases))]
+        public async void ForFriendOrInternalClass_FindsError(
+            string factRelatedAttribute,
+            string classAccessModifier)
```
marcind, line 35: The `classAccessModifier` parameter does not seem to be used.

marcind (changes requested):

test/xunit.analyzers.tests/TestClassMustBePublicTests.cs
wachulski: Overlooked. Fixed now, thx

marcind (approved):

PR #97, opened 2018-02-08
#1559: Update InlineDataMustMatchTheoryParameters with DateTime and Guid con…

wachulski:
…versions.

This addresses conversions introduced in https://github.com/xunit/xunit/pull/1546

Closes xunit/xunit#1559

### DIFF
--- a/src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
+++ b/src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
@@ -16,6 +17,15 @@ namespace Xunit.Analyzers
         internal static readonly string ParameterIndex = "ParameterIndex";
         internal static readonly string ParameterName = "ParameterName";
         internal static readonly string ParameterArrayStyle = "ParameterArrayStyle";
+
+        public InlineDataMustMatchTheoryParameters()
+        {
+        }
+
+        /// <summary>For testing purposes only.</summary>
+        public InlineDataMustMatchTheoryParameters(string assemblyVersion) : base(new Version(assemblyVersion))
+        {
+        }
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
            ImmutableArray.Create(
@@ -109,7 +119,7 @@ namespace Xunit.Analyzers
                         var value = values[valueIdx];
                         if (!value.IsNull)
                         {
-                            var isConvertible = DetermineIsConvertible(compilation, value.Type, parameterType);
+                            var isConvertible = ConversionChecker.IsConvertible(compilation, value.Type, parameterType, xunitContext);
                             if (!isConvertible)
                             {
                                 symbolContext.ReportDiagnostic(Diagnostic.Create(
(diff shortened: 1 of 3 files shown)


src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
```diff
@@ -229,5 +208,65 @@ internal enum ParameterArrayStyleType
...
+                if (destination.TypeKind == TypeKind.TypeParameter)
+                    return IsConvertibleTypeParameter(source, (ITypeParameterSymbol) destination);
+
+                var conversion = compilation.ClassifyConversion(source, destination);
+
+                if (conversion.IsNumeric)
+                    return IsConvertibleNumeric(source, destination);
+
+                // Added in https://github.com/xunit/xunit/pull/1546
+                if (xunitContext.Core.TheorySupportsConversionFromStringToDateTimesAndGuid
```
marcind: Is this correct (i.e. tying the check to xunit's version number)? From https://github.com/xunit/xunit/issues/1494#issue-263912357 it appears that an implicit conversion from string to `DateTime` worked in xunit 2.2.

src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
```diff
@@ -229,5 +208,65 @@ internal enum ParameterArrayStyleType
...
+                    && IsDateTimeRelatedOrGuid(destination))
+                {
+                    // Allow all conversions from strings. Narrowing conversion issues will be reported at runtime.
```
marcind: By "narrowing conversion", do you mean whether the string is actually parsable as a date or guid?

src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
marcind: Looks like https://github.com/xunit/xunit/pull/1546 added support for `DateTimeOffset` and `Guid`, but at least `DateTime` would've worked for earlier version of xunit via [`Convert.ChangeType`](https://docs.microsoft.com/en-us/dotnet/api/system.convert.changetype#System_Convert_ChangeType_System_Object_System_Type_)

marcind (changes requested):
Some questions about support in older versions of xunit.

src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
wachulski: You're completely right. What do you suggest then? I could make a split 2.3+ DateTime and 2.4+ DateTimeOffset + Guid. I was wondering if it's worth it as 2.4 is in beta already. But yes, I admit this would be telling lies and would prevent putting it into 2.3.2 (if happened).

What do you suggest? 

BTW - Have you considered other approaches of xunit exposing its features through some explicit declarations (like assembly attributes etc.). They could be even internal if we made the analyzers its friend. Then the analyzers could use those instead of hardcoding what is possible in version numbers.

src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
wachulski: That's what I meant. I copied the comment from L 259, where it says about numeric conversions. I'll reword it, is `Allow all conversions from strings. All parsing issues will be reported at runtime.` acceptable?

src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
marcind: > I could make a split 2.3+ DateTime and 2.4+ DateTimeOffset + Guid
What do you suggest?

I believe the `DateTime` support goes further back than xunit 2.3, so I would support that unconditionally. The `DateTimeOffset` and `Guid` support needs to be conditional on version 2.4 or higher

> BTW - Have you considered other approaches of xunit exposing its features through some explicit declarations (like assembly attributes etc.).

I want to maintain as loose coupling as possible (and xunit should definitely not know about the analyzers assembly, i.e. making it a friend, plus I'm not sure that would work for whatever context roslyn loads assembly metadata in). There is already conditional support for changes made in xunit 2.3 and earlier and creating updated builds of those earlier versions that include some sort of metadata advertising feature support is not feasible.

src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
marcind: 👍

src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
wachulski: **tl;dr;**
1. Making it friend - I admit this idea is absurd by definition of analyzers nature, the main library should not know anything about analyzers.
2. __I will do DateTime and DateTimeOffset+Guid as you suggested__ thx

The idea of markers is that in general it decouples the versioning out. Imagine the following case: the xunit lib has a PR with feature X, but the analyzers authors don't know when that PR will be merged and under which version it will be released. If the xunit lib marks its assembly with some attribute saying it has feature X on, then the analyzers authors could release code accounting for feature X even if it doesn't exist in the xunit lib yet. They call it __feature toggles__, don't they.

I do also see its cons that one would need to maintain such feature list. It could simply add more clutter and make xunit itself harder to maintain. That's why I'm not convinced myself 😃 **If you find other problems, just challenge me with them, I would be thankful.**

Treat it **just as a general idea** 💡 of how to make analyzers follow the library it analyzes. For sure we need to abide to versioning now (backward compatibility).

wachulski:
I did corrections.

marcind (approved):

PR #99, opened 2018-02-16
Add negative zero as unique value for xUnit1025.

### LINKED ISSUE #1489: Analyzer gives false positive for Zero and minus Zero InlineData
Try adding the following test in Analyzers  InlineDataShouldBeUniqueWithinTheoryTests.ForDuplicatedInlineDataMethod.

     [Fact]
     public async void DoesNotFindWarning_ForZeroAndNegativeZeroInlineData()
     {
           var diagnostics = await CodeAnalyzerHelper.GetDiagnosticsAsync(analyzer,
                    "public class TestClass" +
                    "{" +
                    "   [Xunit.Theory]" +
                    "   [Xunit.InlineData(0.0)]" + Environment.NewLine +
                    "   [Xunit.InlineData(-0.0)]" + Environment.NewLine +
                    "   public void TestMethod(double x) { }" +
                    "}");

                Assert.Empty(diagnostics);
      }

Shouldn't this pass?

wachulski:
Note! This does not cover all possible cases. A trade-off has been made between
algorithm complexity and its ability to catch most common scenarios.
Nested values like (1, new object[] {-0.0}) are disregarded (in this case
it duplicates (1, new object[] {0.0}).

InlineDataShouldBeUniqueWithinTheory analyzer has been refactored.
A common model for argument values created. That led to reduction
of syntax pattern matching - the algorithm got simplified. The common model
has been extracted out to Utilities. Note! Performance was not a subject of
this refactoring, nevertheless ArgumentValue tries to avoid allocations.

Closes xunit/xunit#1489

### DIFF
--- a/src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
+++ b/src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
+using Xunit.Analyzers.Utilities;
 
 namespace Xunit.Analyzers
 {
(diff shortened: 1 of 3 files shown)


marcind (changes requested):
Could you undo the test method rename changes (or at least separate into a different PR)? I prefer to keep functional changes separate from stylistic/cleanup/etc.

wachulski:
Done, as you wished.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
```diff
@@ -76,187 +73,65 @@ private static void ReportDuplicate(SymbolAnalysisContext context, AttributeData
             context.ReportDiagnostic(diagnostic);
         }
 
+        private static AttributeSyntax GetAttributeSyntax(SymbolAnalysisContext context, AttributeData attribute)
```
marcind: Moving this method was unnecessary.

src/xunit.analyzers/Utilities/ArgumentValue.cs
```diff
@@ -0,0 +1,209 @@
...
+                    if (!Values[i].Equals(other.Values[i]))
+                        return false;
+                }
+            }
+            else
+            {
+                return false;
+            }
+
+            return true;
```
marcind, line 153: Isn't this unreachable code?

src/xunit.analyzers/Utilities/ArgumentValue.cs
```diff
@@ -0,0 +1,209 @@
...
+        {
+            if (IsNull && other.IsNull)
+                return true;
+
+            if (!IsArray && !other.IsArray)
+            {
+                if (!Equals(Value, other.Value))
+                    return false;
+
+                // edge case of double/float: -0.0 and 0.0 (we treat as non-equal arguments)
```
marcind: Would be nice to have a reference back to https://github.com/xunit/xunit/issues/1489 here.

src/xunit.analyzers/Utilities/ArgumentValue.cs
```diff
@@ -0,0 +1,209 @@
...
+        {
+            var argumentSyntaxes = (attributeData?
+                                       .ApplicationSyntaxReference.GetSyntax() as AttributeSyntax)?
+                                       .ArgumentList?
+                                       .Arguments ?? new SeparatedSyntaxList<AttributeArgumentSyntax>();
+            int index = 0;
+            var valuesBuilder = ImmutableArray.CreateBuilder<ArgumentValue>();
+            foreach (var typedConstantValue in typedConstants)
+            {
+                var argumentSyntax = index < argumentSyntaxes.Count ? argumentSyntaxes[index] : null;
```
marcind, line 94: `index` is never modified, it's always 0. Probably not the intention.

marcind: I wonder if there's some test coverage missing here for this case.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
```diff
@@ -13,15 +14,16 @@ public class InlineDataShouldBeUniqueWithinTheory : XunitDiagnosticAnalyzer
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
             Descriptors.X1025_InlineDataShouldBeUniqueWithinTheory);
 
-        internal override void AnalyzeCompilation(CompilationStartAnalysisContext compilationStartContext, XunitContext xunitContext)
+        internal override void AnalyzeCompilation(CompilationStartAnalysisContext compilationStartContext,
```
marcind: Unnecessary formatting and whitespace changes in this file.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
```diff
@@ -76,187 +73,65 @@ private static void ReportDuplicate(SymbolAnalysisContext context, AttributeData
...
+
         private static bool HasAttributeDeclarationNoCompilationErrors(AttributeData attribute,
             IArrayTypeSymbol objectArrayType)
         {
             return attribute.ConstructorArguments.Length == 1 &&
                    objectArrayType.Equals(attribute.ConstructorArguments.FirstOrDefault().Type);
         }
 
-        private class InlineDataUniquenessComparer : IEqualityComparer<AttributeData>
+        public class InlineDataUniquenessComparer : IEqualityComparer<AttributeData>
```
marcind: Why the change to `public`?

marcind (changes requested):
I'm a bit torn on this set of changes.

It looks like it achieves the goal of handling simple usages of negative/positive floating 0. As you state, it does not handle more complex (albeit unlikely to be used cases). I'd like to see a unit test for the unhandled case as a form of documentation that this is intentional (I also wonder if the old implementation of the comparison algorithm would handle that case).

As to the rewrite, seems like the goals was to a) simplify the code/reduce duplication and b) optimize performance by reducing boxing and allocations. I'm a little worried here because it looks like there's at least 1 place that introduces a bug (the unchanging `index` variable) and I'm not convinced the perf gains are there. From the perf side, the critical path is the calculation of the hash code for each inline row. I'm not saying the previous version of the code was great at it, but neither is the new one (given the potential for multiple ImmutableArrays being created).

My suggestion would be to try what the -0.0 handling would look like in the existing algorithm, add the tests, and then potentially take a stab at an improved algorithm that takes the performance of the whole analyzer into consideration.

src/xunit.analyzers/Utilities/ArgumentValue.cs
wachulski: I admit, there was a bug and missing coverage indeed. Fixed in ArgumentValue.cs L 96 and unit tests added: DoesNotFind_WhenZeroAndNegativeZeroInlineDataOfDoubleOrFloatAsSecondTestMethodArguments

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
wachulski: I reverted the cleanup, sorry for the inconvenience.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
wachulski: I reverted the cleanup, sorry for the inconvenience.

src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
wachulski: I forgot to make it back private after I had refactored. Done now.

src/xunit.analyzers/Utilities/ArgumentValue.cs
wachulski: Sure thing. In place now.

wachulski:
@marcind 
> It looks like it achieves the goal of handling simple usages of negative/positive floating 0. As you state, it does not handle more complex (albeit unlikely to be used cases). I'd like to see a unit test for the unhandled case as a form of documentation that this is intentional

It was in there from the beginning: src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs L 285

wachulski:
@marcind 
> As to the rewrite, seems like the goals was to a) simplify the code/reduce duplication and b) optimize performance by reducing boxing and allocations.

Well, a) is true but b) was not my intention at all. But I understand you may have thought like that, especially ArgumentValue.cs L 16 states: 

`This data structure should avoid allocations, hence structs (ImmutableArray with a builder) and no LINQ`

Why is it like that? All I wanted was to make the algorithm simpler as it operated on
1. TypedConstants
2. IParamterSymbols
3. (mixed) arrays of the above
That's why I introduced a common model that abstracts away how a test method argument is provided. Please note that I also had to bring argument syntaxes to the algorithm as TypedContants lack syntax info (for -0.0 or 0.0, they just provide the value which is 0.0), so do IParameterSymbols. The first version with pattern matching wasn't as smart as I initially had thought. Now it's simpler, just that. As for a new model I had to make a decision whether it should be a class or struct. I went for structs to avoid allocations.

So, the upshot: I strived for code simplicity (but I admit my remark may have misled you)

wachulski:
> I'm a little worried here because it looks like there's at least 1 place that introduces a bug (the unchanging index variable)

Bug fixed. Good catch thx.

wachulski:
> My suggestion would be to try what the -0.0 handling would look like in the existing algorithm, add the tests

I tried, but it was obvious that the code would get cluttered too much. I refactored first.

This is interesting, since after simplification it occured the previous version contained bugs. Lines:

`xArgPrimitive.Value != yMethodParamDefault.ExplicitDefaultValue` and
`xMethodParamDefault.ExplicitDefaultValue != yArgPrimitive.Value` were simply incorrect, e.g. for

`Finds_WhenDoubleNegativeOrPositiveZerosOfDoubleOrFloatFromParameterDefault` tests failed in cases:
`[InlineData("-0.0", "double")]` and `[InlineData("-0.0f", "float")]` since `(object) 0.0 != (object) 0.0`

Now it got changed to `Equals(Value, other.Value)` (ArgumentValue L 129) and should be ok.

src/xunit.analyzers/Utilities/ArgumentValue.cs
wachulski: Well, this is essential step in this recursive algorithm 🙂 All the preceding steps (except the first one) try to find a single difference which immidiately fails equality. The last one says: *no difference found, so - equal*

wachulski:
@marcind Ping. Any plans about this one?

wachulski:
Closing after a long time of no response/continuation.

PR #103, opened 2018-03-08
Fix assert null on unconstrained generic types

wachulski:
2 commits: actual fix + refactoring (as I already know @marcind likes splitting them)


marcind (approved):
Looks great.

PR #104, opened 2018-03-16
Fix OptionalAttributes included in v2.2 analysis.

wachulski:
BTW - I see `CoreContext.cs` is getting more and more capability markers (`TheorySupports...`). Maybe it's time to make it sth like `XUnitCapabilities` immutable value type created once at the beginning for to pass around?


wachulski:
Got stale

PR #105, opened 2018-04-11
Relax xUnit3001 warning to allow for implicit constructors,

tiesmaster:
fixes [#1672](https://github.com/xunit/xunit/issues/1672).

PR #110, opened 2018-05-14
Remove unused code fixer logic after relaxation of xUnit3001

tiesmaster:
#105 removed the alerting on implicit constructors, so the code fixer doesn't need to generate them anymore, when missing. In #105, I've removed the tests, but failed to update the fixer. This does that.


marcind (approved):
+1

PR #111, opened 2018-05-14
Fix NRE on xUnit1013

### LINKED ISSUE #1694: Analyzer xUnit1013 throws NullReferenceException
repro:
```csharp
using Xunit;

public class C
{
    [Fact]
    public void M()
    {
    }

    public override void M2()
    {
    }
}
```

I did check the source and the problem is that `IMethodSymbol.OverriddenMethod` can return null even if `ISymbol.IsOverride` returns true.

tiesmaster:
This fixes [#1694](https://github.com/xunit/xunit/issues/1694).

WIP because pending discussion on that issue, if this is the right approach.


tiesmaster:
Sorry, forgot to remove my temp commits. I'll rebase, and force push.

PR #112, opened 2018-05-15
Add strong name, Authenticode and NuGet signing

clairernovotny:
Fixes https://github.com/xunit/xunit/issues/1704


marcind:
@onovotny SignClient seems to not be working (e.g. https://ci.appveyor.com/project/xunit/xunit-analyzers/build/0.10.0-build1197). Any idea what might be the problem?

clairernovotny:
There was a hiccup with the signing service, it should be resolved now.

PR #113, opened 2018-05-17
Fix xUnit2000 for named arguments

### LINKED ISSUE #1636: xUnit2000 behavior should account for named arguments
Analyzer version 0.8.0

#### Code:

```csharp
Assert.Equal(actual: int.Parse("3"), expected: 3);
Assert.Equal(actual: 3, expected: int.Parse("3"));
```

#### Expected result:

No warning on the first assertion.
xUnit2000 reported on the second assertion.

#### Actual result:

xUnit2000 is reported on the first assertion, saying the literal should appear first.
No warning on the second assertion.

#### Rationale:

xUnit2000 is intended to catch cases where the *expected* and *actual* values have been unintentionally swapped. Normally this is based on argument positions. However, when named arguments are used, xUnit2000 should analyze the value for `expected` instead of simply looking at the first argument.

tiesmaster:
This fixes [#1636](https://github.com/xunit/xunit/issues/1636). Couple notes:

 * I wanted to handle the edge case where you'd get passed in an invalid named argument (like `exp: `, instead of `expected: `), to avoid NREs from happening. However, I can't get that to fail (both in VS, as well as in unit tests with `compilationReporting` set to `CompilationReporting.IgnoreErrors`), so it seems the analyzers won't be called if those nodes have compilation errors, or so (maybe @sharwell can shine some light on that, since he's on one of the Roslyn teams, and reporter of this issue;) )
 * code fixer needs to be updated, as well, I'll do that in a subsequent PR, if that's ok
 * I've updated the gitignore to account for NCrunch, I hope that's ok

### DIFF
--- a/src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirst.cs
+++ b/src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirst.cs
@@ -1,4 +1,5 @@
 ﻿using System.Threading;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -18,21 +19,31 @@ namespace Xunit.Analyzers
 
         protected override void Analyze(SyntaxNodeAnalysisContext context, InvocationExpressionSyntax invocation, IMethodSymbol method)
         {
-            if (invocation.ArgumentList.Arguments.Count < 2)
+            var arguments = invocation.ArgumentList.Arguments;
+            if (arguments.Count < 2)
                 return;
 
-            var firstArg = invocation.ArgumentList.Arguments[0];
-            var secondArg = invocation.ArgumentList.Arguments[1];
+            ArgumentSyntax expectedArg, actualArg;
+            if (arguments.All(x => x.NameColon != null))
+            {
+                expectedArg = arguments.Single(x => x.NameColon.Name.Identifier.Text == "expected");
+                actualArg = arguments.Single(x => x.NameColon.Name.Identifier.Text == "actual");
+            }
+            else
+            {
+                expectedArg = arguments[0];
+                actualArg = arguments[1];
+            }
 
-            if (IsLiteralOrConstant(secondArg.Expression, context.SemanticModel, context.CancellationToken) &&
-                !IsLiteralOrConstant(firstArg.Expression, context.SemanticModel, context.CancellationToken))
+            if (IsLiteralOrConstant(actualArg.Expression, context.SemanticModel, context.CancellationToken) &&
+                !IsLiteralOrConstant(expectedArg.Expression, context.SemanticModel, context.CancellationToken))
             {
                 var parentMethod = invocation.FirstAncestorOrSelf<MethodDeclarationSyntax>();
                 var parentType = parentMethod.FirstAncestorOrSelf<ClassDeclarationSyntax>();
                 context.ReportDiagnostic(Diagnostic.Create(
                     Descriptors.X2000_AssertEqualLiteralValueShouldBeFirst,
                     invocation.GetLocation(),
-                    secondArg.Expression.ToString(),
+                    actualArg.Expression.ToString(),
                     SymbolDisplay.ToDisplayString(method, SymbolDisplayFormat.CSharpShortErrorMessageFormat.WithGenericsOptions(SymbolDisplayGenericsOptions.None).WithParameterOptions(SymbolDisplayParameterOptions.IncludeName)),
                     parentMethod.Identifier.ValueText,
                     parentType.Identifier.ValueText));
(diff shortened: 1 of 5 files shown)


src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirst.cs
```diff
@@ -18,21 +19,31 @@ public class AssertEqualLiteralValueShouldBeFirst : AssertUsageAnalyzerBase
...
+            var arguments = invocation.ArgumentList.Arguments;
+            if (arguments.Count < 2)
                 return;
 
-            var firstArg = invocation.ArgumentList.Arguments[0];
-            var secondArg = invocation.ArgumentList.Arguments[1];
+            ArgumentSyntax expectedArg, actualArg;
+            if (arguments.All(x => x.NameColon != null))
+            {
+                expectedArg = arguments.Single(x => x.NameColon.Name.Identifier.Text == "expected");
```
sharwell, line 29: :bulb: Use `Identifier.ValueText` to ensure the alternate form `@expected` is handled as well.

src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirst.cs
```diff
@@ -18,21 +19,31 @@ public class AssertEqualLiteralValueShouldBeFirst : AssertUsageAnalyzerBase
...
+            else
+            {
+                expectedArg = arguments[0];
+                actualArg = arguments[1];
+            }
 
-            if (IsLiteralOrConstant(secondArg.Expression, context.SemanticModel, context.CancellationToken) &&
-                !IsLiteralOrConstant(firstArg.Expression, context.SemanticModel, context.CancellationToken))
+            if (IsLiteralOrConstant(actualArg.Expression, context.SemanticModel, context.CancellationToken) &&
+                !IsLiteralOrConstant(expectedArg.Expression, context.SemanticModel, context.CancellationToken))
```
sharwell, line 39: 💭 It seems this would `NullReferenceException` if `exp:` is used instead of `expected:`.

sharwell:
> I wanted to handle the edge case where you'd get passed in an invalid named argument (like exp:, instead of expected:), to avoid NREs from happening.

This analyzer operates at the syntax level, so the argument name will be preserved verbatim in this tree whether or not it's valid. The NRE would occur later. Since the tests are passing, it seems the tests are ignoring AD0001. This is bad - see DotNetAnalyzers/StyleCopAnalyzers@d727e1f8bfceb673fce7a3463440f9d5d7535088 for one example of how to ensure AD0001 always results in a test failure.

src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirst.cs
tiesmaster: Thanks! I always get confused between those two, and usually, they're the same. Good to know a scenario where they differ. I'll fix.

src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirst.cs
tiesmaster: That's the thing, I wanted to cover that, so I put in [a test](https://github.com/xunit/xunit.analyzers/pull/113/files#diff-66f2b6b7a892d46bdf7cb3cabd0a84fdR96) for that, and I just couldn't get it to fail. I also found out that this project actually already handles [AD0001](), but that didn't fire (I sanity checked it by simply forcing a NRE, when the tree is valid, and then I did got that diagnostic as a test failure).

So I wanted to take the host environment out of the equation, and checked in VS (using this [commit](https://github.com/tiesmaster/xunit.analyzers/commit/2b5e10042ce922b9c27b68349d4a115ff15524ea)), with a breakpoint on the `return` (L33), and that just doesn't fire, whatever I type.

~So I thought Roslyn might _short-circuit_ analyzers in certain cases. Doesn't ring a bell? Or maybe a bug? Anyhow, I'm going to browse Roslyn to see if I can find anything.~ So, I went with the Jon Skeet approach, and tried to reproduce the issue in a console app, but couldn't reproduce it, so it's related to this project :(

sharwell: :arrow_right: It's handled here:

https://github.com/xunit/xunit.analyzers/blob/ad63ad341c3cc50c934a62d59da4e551daf236f9/src/xunit.analyzers/Analysis/AssertUsageAnalyzerBase.cs#L42-L44

tiesmaster: Hmm, that's embarrassing. It was right in front of me, all along. I'm not too familiar with this code base yet, but I _did_ take a look at this, but I guessed I just glanced over this 😕 Thanks for your feedback, and time. Next time, I'll take a better look first.

PR #114, opened 2018-05-18
Fix xUnit2000 fixer: support named parameters in swap arguments

tiesmaster:
Follow up on PR #113 to fix the CodeFixer.

### DIFF
--- a/src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirstFixer.cs
+++ b/src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirstFixer.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Immutable;
 using System.Composition;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
@@ -34,10 +35,23 @@ namespace Xunit.Analyzers
         private async Task<Document> SwapArgumentsAsync(Document document, InvocationExpressionSyntax invocation, CancellationToken cancellationToken)
         {
             var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
-            var firstArg = invocation.ArgumentList.Arguments[0];
-            var secondArg = invocation.ArgumentList.Arguments[1];
-            editor.RemoveNode(firstArg);
-            editor.InsertAfter(secondArg, firstArg);
+
+            var arguments = invocation.ArgumentList.Arguments;
+            if (arguments.All(x => x.NameColon != null))
+            {
+                var firstArg = arguments[0];
+                var secondArg = arguments[1];
+
+                editor.ReplaceNode(firstArg, firstArg.WithExpression(secondArg.Expression));
+                editor.ReplaceNode(secondArg, secondArg.WithExpression(firstArg.Expression));
+            }
+            else
+            {
+                var firstArg = arguments[0];
+                var secondArg = arguments[1];
+                editor.RemoveNode(firstArg);
+                editor.InsertAfter(secondArg, firstArg);
+            }
             return editor.GetChangedDocument();
         }
     }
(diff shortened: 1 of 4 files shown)


test/xunit.analyzers.tests/AssertEqualLiteralValueShouldBeFirstFixerTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            var source = string.Format(Template, "Assert.Equal(i, 0)");
+            var expected = string.Format(Template, "Assert.Equal(0, i)");
+
+            var actual = await CodeAnalyzerHelper.GetFixedCodeAsync(analyzer, fixer, source);
+
+            Assert.Equal(expected, actual);
+        }
+
+        [Fact]
+        public async void NamedArgumentsOnlySwapsArgumentValues()
```
marcind, line 34: How about adding a test for the case of named arguments used in the "correct" positions, e.g. `var source = string.Format(Template, "Assert.Equal(expected: i, actual: 0)");`

src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirstFixer.cs
```diff
@@ -34,10 +35,23 @@ public class AssertEqualLiteralValueShouldBeFirstFixer : CodeFixProvider
         private async Task<Document> SwapArgumentsAsync(Document document, InvocationExpressionSyntax invocation, CancellationToken cancellationToken)
         {
             var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
-            var firstArg = invocation.ArgumentList.Arguments[0];
-            var secondArg = invocation.ArgumentList.Arguments[1];
-            editor.RemoveNode(firstArg);
-            editor.InsertAfter(secondArg, firstArg);
+
+            var arguments = invocation.ArgumentList.Arguments;
+            if (arguments.All(x => x.NameColon != null))
```
marcind: C# 7.2 allows the omission of argument names as long as they are positionally in the correct place, e.g. `Assert.Equal(expected: i, 123)`. Certainly an unlikely edge case, but could you make sure that this new syntax is supported?

marcind (changes requested):

test/xunit.analyzers.tests/AssertEqualLiteralValueShouldBeFirstFixerTests.cs
tiesmaster: Nice one. Added.

src/xunit.analyzers/AssertEqualLiteralValueShouldBeFirstFixer.cs
tiesmaster: I actually consider the new syntax, but found out that the order cannot be changed with this, so I thought it wasn't applicable. This is a good scenario, I've covered it now, as well.

test/xunit.analyzers.tests/AssertEqualLiteralValueShouldBeFirstFixerTests.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        [Fact]
+        public async void PartiallyNamedArgumentsInCorrectPositionOnlySwapsArgumentValues()
+        {
+            // C# 7.2 supports this new supported "non-trailing named arguments"
+
+            var source = string.Format(Template, "Assert.Equal(expected: i, 0)");
+            var expected = string.Format(Template, "Assert.Equal(expected: 0, i)");
+
+            var actual = await CodeAnalyzerHelper.GetFixedCodeAsync(analyzer, fixer, source, CompilationReporting.IgnoreErrors);
```
tiesmaster, line 74: I had to ignore compilation errors, since we're still using an older version of Roslyn, without support for the non-trailing arguments. Since updating the dependency can be quite an impact, I figured this was the best way to solve this.

Let me know if you think differently.

marcind: The analyzers should keep the current Roslyn dependency but the tests could target latest (and use compiler settings to target older versions, if necessary). But that's probably a separate change.

marcind (approved):

test/xunit.analyzers.tests/AssertEqualLiteralValueShouldBeFirstFixerTests.cs
tiesmaster: I'll see if I can pick that up.

PR #115, opened 2018-05-24
PR 114 follow up: explicitly specify C# language version

tiesmaster:
As [discussed](https://github.com/xunit/xunit.analyzers/pull/114/files#r190196126), specify C# language version for unit test, since that requires C# 7.2 syntax.

NOTE: I had to also change the target framework for the test project to 4.6, since the latest Roslyn packages are on [netstandard 1.3](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -130,7 +130,7 @@ function __target__signpackages() {
 
 function __target__test() {
     _build_step "Running unit tests"
-        _xunit_console ("test\xunit.analyzers.tests\bin\" + $configuration + "\net452\xunit.analyzers.tests.dll -xml artifacts\test\TestResults.xml -diagnostics")
+        _xunit_console ("test\xunit.analyzers.tests\bin\" + $configuration + "\net46\xunit.analyzers.tests.dll -xml artifacts\test\TestResults.xml -diagnostics")
 }
 
 # Dispatch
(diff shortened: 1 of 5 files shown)


tiesmaster:
@marcind Updating the target framework for the test project to 4.6 is failing the xUnit console runner in the build scripts. Any idea how to fix this?

tiesmaster:
I've spent some time investigating this, but haven't found the actual cause, but could eliminate some things:

 * running the `build.ps` build scripts locally doesn't reproduce the issue for me :(
 * it doesn't seems to be a binding redirect issue, that gets [generated automatically](https://github.com/xunit/xunit.analyzers/blob/master/test/Directory.build.props#L4)
 * this [so post](https://stackoverflow.com/questions/42510558/visual-studio-reports-a-missing-method-in-a-roslyn-analyser-but-it-is-definitely/42511051) talks about the same issue, but downgrading `System.Collections.Immutable` is not allowed by NuGet
 * [other posts](https://stackoverflow.com/questions/948785/typeloadexception-says-no-implementation-but-it-is-implemented/948787#948787) talk, _in general_, about "System.TypeLoadException: Method 'get_xxx' does not have an implementation", but can't figure out what would exactly happen in our case, and why it happens only on AppVeyor, and most importantly, how to fix it.

tiesmaster:
So after a bit of RDP'ing into AppVeyor, and comparing that environment to mine, I've figured out the difference. On AppVeyor, `System.Collections.Immutable` (v1.1.37) is installed in the GAC, and that causes this problem. When I also install this version locally in the GAC, the problem reproduces.

I don't understand completely why having this assembly in the GAC results in this error, but I did found out that denying appdomains to the runner causes the binding redirects to not take effect. When I allow appdomains, the `app.config` gets picked up, and the problem goes away.

@marcind I see that you've made this change (75d0610) to speed up the tests, but it's only 2 seconds, so the gain is not that much, IMHO. Are you ok with changing it back?

tiesmaster:
@marcind Did you had a time to look at this PR? Do you have any concerns with this, or any questions?

PR #118, opened 2018-07-03
Add xUnit1027: Collection definition classes must be public

josetr:
Right now If you forget to make it public you get the error **The following constructor parameters did not have matching fixture data: FixtureClass fixture**. 

The problem is that this error is shown at runtime and doesn't really tell you where the problem lies and how to fix it. How long it takes you to find the missing piece depends on your xunit knowledge. In my case, it took me 20 minutes because I made this pull request 😄

### DIFF
--- a/docs/_rules/xUnit1027.md
+++ b/docs/_rules/xUnit1027.md
@@ -1,0 +1,40 @@
+---
+title: xUnit1027
+description: Collection definition classes must be public
+category: Usage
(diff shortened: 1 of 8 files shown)


bradwilson:
Thanks! (Sorry for the delay.) Since the source has been significantly overhauled, I have ported your fixes here (d35f71b) and pushed them up.

PR #121, opened 2019-06-05
`Assert.NotNull({{expr1}.FirstOrDefault({expr2}))` -> `Assert.Contains({expr1}, {expr2})` and others

dnikolovv:
Implements issue [1508](https://github.com/xunit/xunit/issues/1508). If that's OK I could also tackle [1510](https://github.com/xunit/xunit/issues/1510) and [1511](https://github.com/xunit/xunit/issues/1511).

Behavior:

`Assert.Null({expr}.FirstOrDefault())` -> `Assert.Empty({expr})`
`Assert.NotNull({expr}.FirstOrDefault())` -> `Assert.NotEmpty({expr})`
`Assert.Null({expr}.FirstOrDefault({expr1}))` -> `Assert.DoesNotContain({expr}, {expr1})`
`Assert.NotNull({expr}.FirstOrDefault({expr1}))` -> `Assert.Contains({expr}, {expr1})`

I've yet to come up with a better name for the rule. Any suggestions?

PS. The failing build seems to result from an issue that occurred more than 5 months ago - https://ci.appveyor.com/project/xunit/xunit-analyzers/builds/21272768


markusschaber:
> `Assert.Null({expr}.FirstOrDefault())` -> `Assert.Empty({expr})`
> `Assert.NotNull({expr}.FirstOrDefault())` -> `Assert.NotEmpty({expr})`

Those expressions are not equal when the sequence `expr` conains a `null` value as first element.

> `Assert.Null({expr}.FirstOrDefault({expr1}))` -> `Assert.DoesNotContain({expr}, {expr1})`
> `Assert.NotNull({expr}.FirstOrDefault({expr1}))` -> `Assert.Contains({expr}, {expr1})`

Similar here, when the sequence contains a `null` value which satisfies the condition `expr1`.

Those might be corner cases, but it should be clear that it's changing semantics.

The original examples in https://github.com/xunit/xunit/issues/1508 differ in the way that they use the `.OfType<T>()` extension method which never returns null, so the sequence `expr` is guaranteed to never contain `null`s.

dnikolovv:
@markusschaber You're right. Do you think we can somehow make it clear?

markusschaber:
@dnikolovv I'm not sure. Maybe the analysis should be refined and the warning restricted to those cases where it can be deduced that the sequence will not contain `null`s, as it's in the original example. (This analysis might be easier with C# 8.0).

Explaining the corner cases as an exception to the warning makes it difficult to understand, and also forces the user to suppress the warnings in those cases.

bradwilson:
Sorry for the delay getting back to this.

The more I think about it, I think I'd rather not take this. Because it's changing the semantic meaning to something that's not 100% equivalent, I don't think we should make this recommendation. It's not a case of the code being wrong, but rather a case of the code being _potentially_ harder to read than it needs to be. In general, I would be okay with making code easier to read (especially to introduce developers to assertions they may not realize exist), but in this case--because we can't guarantee that the fixed code is always identical to the original code--I think we're better off just not doing it.

PR #123, opened 2019-06-17
Split Code Fixes to their own assembly

sharwell:
The compiler does not provide the workspaces layer during command line builds. Removing the reference to the workspaces layer from the analyzer assembly ensures that the analyzers will never silently fail to load due to an accidental reference to the workspaces layer.

:warning: To avoid unnecessary headaches in follow-up pull requests, please do not rebase or squash this pull request during a merge.


sharwell:
@marcind This is now ready. Please note the request to _not_ rebase or squash this during the merge. Each commit was intentionally a distinct and complete change.

global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "2.1.504"
```
bradwilson, line 3: Why does this not work with later versions of the SDK?

Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion>latest</LangVersion>
+    <LangVersion>7.3</LangVersion>
```
bradwilson, line 4: Why does this not work with the latest version of C#?

src/xunit.analyzers/xunit.analyzers.csproj
```diff
@@ -17,12 +17,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="1.2.0-beta2" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="1.2.2" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <Analyzer Include="$(NuGetPackageRoot)microsoft.codeanalysis.analyzers\1.2.0-beta2\analyzers\dotnet\cs\Microsoft.CodeAnalysis.Analyzers.dll" />
-    <Analyzer Include="$(NuGetPackageRoot)microsoft.codeanalysis.analyzers\1.2.0-beta2\analyzers\dotnet\cs\Microsoft.CodeAnalysis.CSharp.Analyzers.dll" />
```
bradwilson, line 25: Why did these get deleted?

sharwell: ➡️ They are automatically added by `<PackageReference>`.

Directory.Build.props
sharwell: ➡️ The "latest" version of C# isn't a version and should never be used. 7.3 declares a specific version that will work consistently for developers using Visual Studio 2017 version 15.7 and higher. It avoids situations where the IDE and build behaves differently (read: not predictably) for some users.

global.json
sharwell: Some of the NuGet dependencies were getting resolved differently and it wouldn't complete a build. You can see the long tail of failed AppVeyor builds for examples. Even a trivial no-op commit will trigger a build failure in the current state of the repository.

After this change is merged, we can go back and work to update the SDK version if we want. This at least gets us to a state where builds work though.

bradwilson:
There are a ton of package version downgrade warnings when building this. See https://ci.appveyor.com/project/xunit/xunit-analyzers/builds/26671167 lines 334-450.

bradwilson:
There are a lot of very outdated packages, too. I'm going to see if getting everything up to date makes the build happier.

sharwell:
@bradwilson I already have the updates well underway. I'll send the next piece before Labor Day.

PR #124, opened 2019-07-17
Closes xunit/xunit#1877 and fixes for surfaced bugs after the first fix

### LINKED ISSUE #1877: Invalid xUnit1025 diagnostic for different InlineData values
```c#
[Theory]
[InlineData(new int[] {1, 2, 3}, new int[] {},        new int[] {},        new int[] {1, 2, 3})]
[InlineData(new int[] {1, 2, 3}, new int[] {1, 2, 3}, new int[] {},        new int[] {})] //      <-- this line
public void ProducesCorrectDiff(int[] before, int[] after, int[] added, int[] removed)
```

The analyzer thinks the highlighted `InlineData` is a duplicate which it clearly isn't.

**EDIT** given improved understanding of the issue.

laurihelkkula:
Closes xunit/xunit#1877. This was a fairly straightforward fix.

But the fix made the test [FindsError_WhenArgumentsAreArrayOfValuesAndTestMethodOffersDefaultParameterValues ](https://github.com/xunit/xunit.analyzers/blob/cebd67a5c3534e43a2e7b5fa0367d9ed1260bb3c/test/xunit.analyzers.tests/InlineDataShouldBeUniqueWithinTheoryTests.cs#L264) fail. The reason turned out to be a different bug that was hidden by the first bug, and was found by a little luck. The root cause for that was object reference comparison of boxed values, and while there was one test case for it, [FindsError_WhenDuplicatedByDefaultValueOfParameter](https://github.com/xunit/xunit.analyzers/blob/cebd67a5c3534e43a2e7b5fa0367d9ed1260bb3c/test/xunit.analyzers.tests/InlineDataShouldBeUniqueWithinTheoryTests.cs#L282), it used the value 1 as the default parameter value. That value did not trigger the lurking bug for some reason, but other values, such as 40 (that was used in FindsError_WhenArgumentsAreArrayOfValuesAndTestMethodOffersDefaultParameterValues), and 2, would trigger it. There were actually three similar bugs in that method. I added test cases for all of them.

Since this is my first contribution here, there might be something that I have overlooked. Let me know if there is something further to do here.

### DIFF
--- a/src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
+++ b/src/xunit.analyzers/InlineDataShouldBeUniqueWithinTheory.cs
@@ -130,7 +130,7 @@ namespace Xunit.Analyzers
                                         return false;
                                     break;
                                 case IParameterSymbol yMethodParamDefault:
-                                    if (xArgPrimitive.Value != yMethodParamDefault.ExplicitDefaultValue)
+                                    if (!object.Equals(xArgPrimitive.Value, yMethodParamDefault.ExplicitDefaultValue))
                                         return false;
                                     break;
                                 default:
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks! (And sorry for the delay) Since the testing framework for analyzers has changed, it was simpler for me to port your code and commit it here (751008e) so I'll close this rather than merging it.

PR #126, opened 2019-08-28
check for ConfigureAwaited tasks

SimonCropp:
so we get a warning for

```
class TestClass { void TestMethod() {
    Xunit.Assert.Throws<System.ArgumentException>(
"param1",
() => System.Threading.Tasks.Task.Delay(0));
} }
```


bradwilson:
Thanks!

PR #129, opened 2019-09-16
Prepare analyzer assemblies for C# and Visual Basic

sharwell:
This pull request is broken into several commits, each of which is expected to build and test successfully.

📝 Please do not rebase or squash this pull request during the merge. I'm putting significant effort into the separation by commit, with the goal of allowing the changes to be easily referenced or reviewed in the future both within this project and by members of other analyzer projects hoping to use this as an example.


ivanbasov:
> namespace Xunit.Analyzers

namespace does not match the folder

---
Refers to: src/xunit.analyzers.csharp/Analysis/AssertUsageAnalyzerBase.cs:9 in f9beee0. [](commit_id = f9beee02c903b2ff59cf784466f8907cc5c0c7b4, deletion_comment = False)

ivanbasov:
> namespace Xunit.Analyzers.Utilities

namespace does not match the folder

---
Refers to: src/xunit.analyzers.csharp/Utilities/TypeHierarchyComparer.cs:5 in f9beee0. [](commit_id = f9beee02c903b2ff59cf784466f8907cc5c0c7b4, deletion_comment = False)

ivanbasov (approved):
:shipit:

ivanbasov:
> /cc @jasonmalinowski @ivanbasov

Thank you, @sharwell ! Great work!

sharwell:
@ivanbasov The files were renamed without any content changes. GitHub shows this during review. If your code review tool does not, I suggest filing a bug with the maintainer of the review tool to get that corrected.

ivanbasov:
> @ivanbasov The files were renamed without any content changes. GitHub shows this during review. If your code review tool does not, I suggest filing a bug with the maintainer of the review tool to get that corrected.

The current folder is xunit.analyzers.csharp. One could assume that there would be another folder xunit.analyzers.vb. Then, it would be great to separate files in those folder with namespaces. If we do not expect to have xunit.analyzers.vb, then we may remove csharp from the current folder as well.

bradwilson:
There are no plans to support VB.

sharwell:
@bradwilson We were hoping to have the analyzers (or a usable subset like seen in this pull request) support VB by volunteering the implementation and tests. Would that be an acceptable strategy?

bradwilson:
The problem is that partial support creates a situation where people are confused why only a few of the advertised analyzers are available. Even if you implement them all it still ends up placing a burden on us to write new analyzers and fixers for VB.

We have received zero requests to support VB.

sharwell:
@bradwilson I have received several requests for this feature (two of them were referenced in https://github.com/xunit/xunit.analyzers/pull/129#issuecomment-531898885). I decided to implement it rather than forward the request in order to expedite delivery of it.

bradwilson:
We're not supporting VB analyzers. This is not open to debate. Sorry.

PR #131, opened 2019-10-17
Allowing analyzers to pass derived Xunit attributes

DillonAd:
This pull request is to address [#1963](https://github.com/xunit/xunit/issues/1963) by adding a use case to the TheoryMethodMustHaveTestData analyzer to allow attributes that inherit from the Xunit.TheoryAttribute class.

Excited and ready for feedback!

### DIFF
--- a/src/xunit.analyzers/TheoryMethodMustHaveTestData.cs
+++ b/src/xunit.analyzers/TheoryMethodMustHaveTestData.cs
@@ -17,8 +17,10 @@ namespace Xunit.Analyzers
             {
                 var symbol = (IMethodSymbol)symbolContext.Symbol;
                 var attributes = symbol.GetAttributes();
-                if (attributes.ContainsAttributeType(xunitContext.Core.TheoryAttributeType) &&
-                    (attributes.Length == 1 || !attributes.ContainsAttributeType(xunitContext.Core.DataAttributeType)))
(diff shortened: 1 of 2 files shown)


bradwilson:
Sorry for the delay in getting to this.

Thinking about this, I'd like to understand your use case here. The reason we were explicit about it is that we assume that anybody which is going to derive from `FactAttribute` (or `TheoryAttribute`) is because they want to change the behavior, and they would be adding a new test case discoverer. This means we can't realistically make any assumptions that we understand how that new attribute works or what it expects for data.

Part of me thinks that the right behavior here is to actually seal `TheoryAttribute` because if you are deriving from it for the purposes of changing its behavior, then you may as well have derived from `FactAttribute` instead. The actual attribute class contains no logic of its own.

powercode:
A `RunnableOnlyInDebuggerTheoryAttribute` would be one such use case, where it skips the test if it is not running in a debugger.

bradwilson:
In retrospect, I'm going to close this because I'm not convinced that my take isn't correct, and this analyzer would lead to potentially incorrect analysis with no way for the user to opt out of that analysis when it's incorrect (short of suppression). In addition, I am strongly considering sealing `TheoryAttribute` in v3 for the reason I stated about.

PR #132, opened 2019-10-22
remove redundant 'as' checks

SimonCropp:
Since Analyzers are filtered by node type. the `as` type checks are redundant


bradwilson:
Thanks!

PR #137, opened 2020-02-26
Fix for false positive of xUnit1026 when an out parameter is used for Moq setup (#1802)

FabianNitsche:
Fix for issue https://github.com/xunit/xunit/issues/1802
It basically not only looks for read parameters in the code analysis but also respects captured ones. Since Moq is using this capturing to actually read the value.


FabianNitsche:
The checks failed because of some slack notification. What can I do?

bradwilson:
@FabianNitsche I fixed the Slack notification issue and merged in master, and the checks should pass now. I'll merge the PR shortly.

bradwilson:
Thanks!

PR #138, opened 2020-03-01
Allow generic type arrays in Theories

### LINKED ISSUE #2075: Analyzer fires xUnit1010 for valid Theory with generic array provided by InlineData
I'm writing generic delimitted-string parsing function, and wish to test it using a Theory and InlineData.

Here's a very stripped-down example of the code and its test:

```csharp
public class UnitTest1
{
    [Theory]
    [InlineData("1,2", new[] { 1, 2 })]
    [InlineData("false,true", new[] { false, true })]
    [InlineData("x,y", new[] { "x", "y" })]
    public void Test1<T>(string input, T[] expected)
        => Assert.Equal(expected, Parse<T>(input));

    private static IEnumerable<T> Parse<T>(string input)
        => input.Split(',').Select(s => (T)Convert.ChangeType(s, typeof(T)));
}
```

The xUnit analyzer currently flags the array parameter as invalid:

image

However, if the warning is suppressed, the tests do execute as desired:

image

Since the runner can execute the tests, this appears to be a bug in the analyzers.

alexbestul:
Fixes xunit/xunit#2075

### DIFF
--- a/src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
+++ b/src/xunit.analyzers/InlineDataMustMatchTheoryParameters.cs
@@ -215,6 +215,14 @@ namespace Xunit.Analyzers
 		{
 			public static bool IsConvertible(Compilation compilation, ITypeSymbol source, ITypeSymbol destination, XunitContext xunitContext)
 			{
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks!

PR #139, opened 2020-04-19
Fix #1968: Analyzer should attempt to discover whether any member would actually be valid

### LINKED ISSUE #1968: Analyzer xUnit1016 / xUnit1019 fired even if matching method
When test data generation methods (targets of `MemberData`) are considered by analyzers, it appears to only look at the _first_ method encountered, rather than attempting to discover whether any would actually be valid.  This appears to occur even if the method is completely invalid, due to things like parameter lists.

image

Switching the order of the public and private method here quiets the analyzer.
Note that, prior to the analyzer being enabled, the code would successfully build and run tests.

[Attaching the sample repro](https://github.com/xunit/xunit/files/3308756/PublicMemberReturnType.zip)

zaytsev-victor:
Fixes xunit/xunit#1968

### DIFF
--- a/src/xunit.analyzers/MemberDataShouldReferenceValidMember.cs
+++ b/src/xunit.analyzers/MemberDataShouldReferenceValidMember.cs
@@ -176,24 +176,36 @@ namespace Xunit.Analyzers
 
 		static ISymbol FindMemberSymbol(string memberName, ITypeSymbol type, int paramsCount)
 		{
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks! Due to merge conflicts, I ported your code here (e0f38062efcced684d717ba661578f16c675913b) and merged it.

PR #149, opened 2022-07-29
Update AssertNullShouldNotBeCalledOnValueType and AssertSameShouldNotBeCalledOnValueTypes to not trigger on arguments containing user-defined implicit conversions

### LINKED ISSUE #2395: xUnit2002 flags a literal int that's being cast to a reference type as not nullable
Build the following project which does this:
```c#
Assert.NotNull((JsonValue?)(int?)42);
```

The cast here should treat that `int` as a nullable reference type `JsonValue`.  It's possible for a badly implemented conversion operator to return `null` here, which is what we're testing.  The analyzer thinks we're checking the integer for `null`.

```
C:\scratch\xunitAssert\TestClass.cs(13,13): warning xUnit2002: Do not use Assert.NotNull() on value type 'int'. Remove this assert. [C:\scratch\xunitAssert\xunitAssert.csproj]
```

If you examine the IL of this method you can see it is indeed calling the operator and passing the result into `Assert.Equals(object)`, so there seems to be a bug in how the analyzer identifies the type passed in.  It could be related to implicit casting that happens.  To get from int -> JsonValue the compiler discovers the implicit cast to JsonNode.  I've recreated an isolated repro here that doesn't use the Json types: 
[xunitAssert.zip](https://github.com/xunit/xunit/files/7289721/xunitAssert.zip)

bjornhellander:
Fixes https://github.com/xunit/xunit/issues/2395

The old code in WalkDownImplicitConversions checked `IOperation.IsImplicit` instead of `IConversionOperation.Conversion.IsImplicit`, which apparently was not the same thing.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/AssertNullShouldNotBeCalledOnValueTypesTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/AssertNullShouldNotBeCalledOnValueTypesTests.cs
@@ -108,4 +108,36 @@ class Class<T> {{
 
 		await Verify.VerifyAnalyzerAsyncV2(source);
 	}
+
+	[Theory]
+	[MemberData(nameof(Methods))]
+	// https://github.com/xunit/xunit/issues/2395
+	public async void DoesNotFindWarning_ForUserDefinedImplicitConversion(string method)
+	{
+		var source = $@"
+public class TestClass
+{{
+    public void TestMethod()
+    {{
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/xunit/xunit.analyzers?pullRequest=149) <br/>All CLA requirements met.

bradwilson:
Thanks!

PR #150, opened 2022-08-04
Fix xUnit1033 when constructor has more parameters than fixture

bdovaz:


bdovaz:
I haven't commented on it but I had doubts whether that first check of the number of parameters still makes sense.

bradwilson:
Thanks!

PR #152, opened 2022-09-11
xUnit2005 not triggered when one parameter is null

### LINKED ISSUE #2560: xUnit2005 not triggered when one parameter is null
Repro code:
```
Assert.Same(17, 18);
Assert.Same(17, null);
Assert.Same(null, 17);
```

Expected result:
xUnit2005 triggered for all asserts, since at least one of the values are of value type.

Actual result:
xUnit2005 only triggered for the first assert. (xUnit2003 is however triggered for the last one.)

xunit version: 2.4.1

Sevarius:
This PR fix [issue 2560](https://github.com/xunit/xunit/issues/2560).


bradwilson:
Thanks!

PR #154, opened 2023-03-03
InlineDataShouldBeUniqueWithinTheory: Treat new array and null as unique

### LINKED ISSUE #2662: False warning of duplicate InlineData, although there is no duplication
Below code shows a simple unit test which has two inline datas 
- one case is to run the test with NULL 
- another case is to run the test with an array containing a single value 0.

The test itself is meaningless but it is sufficient to repro the issue.

```CSHARP
    public class UnitTest1
    {
        [Theory]
        [InlineData(new[] { 0 })]
        [InlineData(null)]
        public void Demo_Test(int[] arr)
        {
            Assert.True(true);
        }
    }
```

**Steps to reproduce:**
- Create an xUnit test project using Visual Studio targeting .NET 6
- Add the test provided above
- Build the project and check the warnings

**Expected** - There should be no warnings as there are no duplicate inline data attributes.

**Actual** - It seems that if a theory is written as mentioned below, we get a warning saying 

> Warning-AD0001 **Analyzer 'Xunit.Analyzers.InlineDataShouldBeUniqueWithinTheory' threw an exception of type 'System.InvalidOperationException' with message 'This operation cannot be performed on a default instance of ImmutableArray<T>.  Consider initializing the array, or checking the ImmutableArray<T>.IsDefault property.'.**	

The image given below shows the package versions used by the xUnit test project.. 
image

ghost:
Closes https://github.com/xunit/xunit/issues/2662

Basically, we already determined by this point that both weren't null by inline data or default param by the caller, but we have yet to determine if one is, but not the other. If x is, but y isn't, everything is fine because the outer switch case checks that. The problem is if x isn't, but y is because the inner switch wasn't checking this so if that case happened, we get a default ImmutableArray when doing the Cast which throws and it reported it as a warning.

Adding a check prevents this and allows the analyser to conclude they are different.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/InlineDataShouldBeUniqueWithinTheoryTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/InlineDataShouldBeUniqueWithinTheoryTests.cs
@@ -152,6 +152,20 @@ public class TestClass {
 		}
 
 		[Fact]
+		public async void DoesNotFindError_WhenNewArrayAndNullDataAttributes()
+		{
+			var source = @"
+public class TestClass{
+    [Xunit.Theory]
+    [Xunit.InlineData(new[] { 0 })]
+    [Xunit.InlineData(null)]
+    public void TestMethod(int[] arr) { }
+}";
+
+			await Verify.VerifyAnalyzerAsyncV2(source);
+		}
+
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks!

PR #155, opened 2023-03-03
AssertEqualLiteralValueShouldBeFirst: Treat a string constructor with constants as a constant

### LINKED ISSUE #2394: xUnit2000 analyzer doesn't treat call to string constructor as a literal
Consider a test which wishes to check that a public const does not change.

The API being tested is https://github.com/dotnet/runtime/blob/355a8cf12b21d81248aa8f5fc7ca2544e4a63db5/src/libraries/System.Private.CoreLib/src/System/CodeDom/Compiler/IndentedTextWriter.cs#L20
```
public const string DefaultTabString = "    ";
```

The test was testing as https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.IO/tests/IndentedTextWriter.cs#L315
```
Assert.Equal(new string(' ', 4), IndentedTextWriter.DefaultTabString);
```

We see the following diagnostic with the latest analyzers
```
The literal or constant value IndentedTextWriter.DefaultTabString should be passed as the 'expected' argument in the call to 'Assert.Equal(expected, actual)' in method 'Ctor_ExpectedDefaults' on type 'IndentedTextWriterTests'. Swap the parameter values.
```

We worked around this by changing the expected value from `new string(' ', 4)` to `"    "` but it seems a bit less readable.  It seems possible that this analyzer could see the call to the string constructor with constants as a constant.

ghost:
… constants as a constant

Closes https://github.com/xunit/xunit/issues/2394

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/AssertEqualLiteralValueShouldBeFirstTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/AssertEqualLiteralValueShouldBeFirstTests.cs
@@ -46,6 +46,19 @@ class TestClass {{
 		await Verify.VerifyAnalyzerAsyncV2(source);
 	}
 
+	[Fact]
+	public async void DoesNotFindWarningForConstantsUsedInStringConstructorAsFirstArgument()
+	{
+		var source = @"
+class TestClass {
+    void TestMethod() {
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks!

PR #158, opened 2023-07-15
2749 - Restrict return type check to either Fact or Theory

### LINKED ISSUE #2749: Update xUnit1028 to only trigger on Fact and Theory
@ploeh wrote in #2748:

> May I suggest that the analyser only considers [Fact] and [Theory], but not third-party attributes that derive from FactAttribute?

Vannevelj:
Fixes https://github.com/xunit/xunit/issues/2749


Vannevelj:
@dotnet-policy-service agree

bradwilson:
Thanks!

PR #164, opened 2023-10-31
[xUnit2189] Extend xUnit1012 to test reference type nullability for InlineDataAttribute

JamesTerwilliger:
Partial fix for this issue:
https://github.com/xunit/xunit/issues/2189

Extends the analyzer and fixer for xUnit1012 to checking null literals against nullability of parameter types for InlineDataAttribute. Without this change, null arguments in InlineDataAttribute are only flagged if they are to be passed to a method parameter that is value-typed and non-nullable. With this change, the analyzer also checks for nullable/non-nullable reference types, and allows the fixer to adjust it as well.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/InlineDataMustMatchTheoryParametersTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/InlineDataMustMatchTheoryParametersTests.cs
@@ -1355,6 +1355,99 @@ public class TestClass {{
 			await Verify.VerifyAnalyzer(source);
 		}
 
+		[Theory]
+		[InlineData("object")]
+		[InlineData("string")]
+		[InlineData("System.Exception")]
+		public async void NonNullableReferenceTypes(string type)
+		{
+			var source = $@"
+#nullable enable
+public class TestClass {{
+    [Xunit.Theory]
+    [Xunit.InlineData(1, null)]
+    public void TestMethod(int a, {type} b) {{ }}
+#nullable restore
+}}";
+
+			DiagnosticResult[] expected =
+			{
+				Verify
+					.Diagnostic("xUnit1012")
+					.WithSpan(5, 26, 5, 30)
+					.WithSeverity(DiagnosticSeverity.Warning)
+					.WithArguments("b", type)
+			};
+
+			await Verify.VerifyAnalyzer(Microsoft.CodeAnalysis.CSharp.LanguageVersion.CSharp8, source, expected);
+		}
+
+		[Theory]
+		[InlineData("object")]
+		[InlineData("string")]
+		[InlineData("System.Exception")]
+		public async void NullableReferenceTypes(string type)
+		{
+			var source = $@"
+#nullable enable
+public class TestClass {{
+    [Xunit.Theory]
+    [Xunit.InlineData(1, null)]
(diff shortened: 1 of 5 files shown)


src/xunit.analyzers/Utility/Descriptors.cs
```diff
@@ -146,10 +146,10 @@ public static class Descriptors
 	public static DiagnosticDescriptor X1012_InlineDataMustMatchTheoryParameters_NullShouldNotBeUsedForIncompatibleParameter { get; } =
 		Rule(
 			"xUnit1012",
-			"Null should not be used for value type parameters",
+			"Null should not be used for value type or non-nullable reference type parameters",
```
bradwilson: Hmm, I'm trying to think of slightly less awkward wording that avoids the "or".

`Null should only be used for nullable parameters.`
`Null should only be used for parameters of nullable types.`

Thoughts?

src/xunit.analyzers.tests/Analyzers/X1000/InlineDataMustMatchTheoryParametersTests.cs
```diff
@@ -1355,6 +1355,99 @@ public class TestClass {{
...
+			DiagnosticResult[] expected =
+			{
+				Verify
+					.Diagnostic("xUnit1012")
+					.WithSpan(5, 26, 5, 30)
+					.WithSeverity(DiagnosticSeverity.Warning)
+					.WithArguments("b", type)
+			};
+
+			await Verify.VerifyAnalyzer(Microsoft.CodeAnalysis.CSharp.LanguageVersion.CSharp8, source, expected);
```
bradwilson: These can be simplified with a using for `Microsoft.CodeAnalysis.CSharp`.

JamesTerwilliger:
@bradwilson I have part of the MemberDataAttribute analyzer done as well - perhaps I should check that in separately for the sake of smaller PRs?

bradwilson:
That's up to you. I'm happy to take it either way. 😄

JamesTerwilliger:
Let's do them separately then, I like smaller PRs with singular focus when possible. PR comments resolved so far.

bradwilson:
Based on https://github.com/xunit/xunit/issues/2808 I'm almost certainly going to need to ship a new core framework build (2.6.1), so I'd like to wait for both of these PRs to merge so that I can also ship a new analyzer's build.

Unless you have more planned than just this fix... 😁

bradwilson:
Thanks!

JamesTerwilliger:
> Based on [xunit/xunit#2808](https://github.com/xunit/xunit/issues/2808) I'm almost certainly going to need to ship a new core framework build (2.6.1), so I'd like to wait for both of these PRs to merge so that I can also ship a new analyzer's build.
> 
> Unless you have more planned than just this fix... 😁

At the moment, what I have planned for the second PR is:
- Detect if arguments to MemberDataAttribute actually match the parameters to the method
- Detect if the return type of the method allows nulls and if so flag the test method parameters that are non-nullable (I have this one ready with some tests and can check it in by itself)

I also noticed that with the v3 change I made to allow MemberDataAttribute to work with optional parameters, the analyzer may report that a method cannot be found to match the arguments even if there is now one that does match. I was planning on doing that as well this week.

PR #165, opened 2023-11-01
[xUnit2189] Add analyzers to test for MemberDataAttribute argument mismatches with method

JamesTerwilliger:
This PR adds analyzers for the following cases:
- MemberDataAttribute references a method that returns null values, but the theory method has non-nullable parameters
- MemberDataAttribute has arguments that are null, but the corresponding parameter in the method is non-nullable
- MemberDataAttribute has arguments that are not type-compatible with the corresponding method parameter
- MemberDataAttribute has more arguments than the referenced method has parameters

The PR intends to fix the remainder of https://github.com/xunit/xunit/issues/2189 plus some other cases that were marked as TODO in the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,22 +31,19 @@ If you installed xUnit.net v2 2.3.0 or higher and do not wish to use the analyze
     <tr>
       <th></th>
       <th>Stable</th>
-      <th>Latest prerelease</th>
       <th>Latest CI</th>
-      <th>CI status</th>
+      <th>Build status</th>
     </tr>
   </thead>
   <tbody>
     <tr>
       <th>v2 core framework</th>
       <td><a href="https://www.nuget.org/packages/xunit"><img src="https://img.shields.io/nuget/v/xunit.svg?logo=nuget"></a></td>
-      <td><a href="https://www.nuget.org/packages/xunit/absoluteLatest"><img src="https://img.shields.io/nuget/vpre/xunit.svg?logo=nuget"></td>
       <td><a href="https://feedz.io/org/xunit/repository/xunit/packages/xunit"><img src="https://img.shields.io/badge/endpoint.svg?url=https%3A%2F%2Ff.feedz.io%2Fxunit%2Fxunit%2Fshield%2Fxunit%2Flatest"></td>
       <td><a href="https://actions-badge.atrox.dev/xunit/xunit/goto?ref=v2"><img src="https://img.shields.io/endpoint.svg?url=https%3A%2F%2Factions-badge.atrox.dev%2Fxunit%2Fxunit%2Fbadge%3Fref%3Dv2&amp;label=build"></td>
     <tr>
     <tr>
       <th>v3 core framework (<a href="https://xunit.net/docs/v3-alpha">alpha docs</a>)</th>
-      <td>N/A</td>
       <td>N/A</td>
       <td><a href="https://feedz.io/org/xunit/repository/xunit/packages/xunit.v3"><img src="https://img.shields.io/badge/endpoint.svg?url=https%3A%2F%2Ff.feedz.io%2Fxunit%2Fxunit%2Fshield%2Fxunit.v3%2Flatest"></td>
       <td><a href="https://actions-badge.atrox.dev/xunit/xunit/goto?ref=main"><img src="https://img.shields.io/endpoint.svg?url=https%3A%2F%2Factions-badge.atrox.dev%2Fxunit%2Fxunit%2Fbadge%3Fref%3Dmain&amp;label=build"></td>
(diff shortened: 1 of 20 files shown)


JamesTerwilliger:
I don't have fixers for these yet - I may give that a try as a separate checkin. These fixers would be harder to do implement than their InlineData counterparts.

JamesTerwilliger:
Update: Two fixers added, probably the easiest ones

bradwilson:
Started reviewing the changes.

First: I think this is big enough that I don't want to delay 2.6.1. I'll push a separate analyzers release later.

Second: Looking at `object[]` vs. `object?[]` as the "hint" for nullability is a tough one, because you don't know which values might be null and which ones won't, so forcing them to mark all parameters as nullable feels like overkill. I tend to want to prefer things which are cut & dried true or false, or else developers will just end up turning off the rule entirely in frustration.

Related, though: we offer `TheoryData<>` which has a much stronger type guarantee. You'd have to look for it explicitly, because it boils down to `IEnumerable<object[]>` at the bottom, but you could use the generic arguments to it for nullability. That's a much stronger guarantee (because it's done by the compiler) and you don't have to guess which elements might be nullable or not.

JamesTerwilliger:
> Second: Looking at `object[]` vs. `object?[]` as the "hint" for nullability is a tough one, because you don't know which values might be null and which ones won't, so forcing them to mark all parameters as nullable feels like overkill. I tend to want to prefer things which are cut & dried true or false, or else developers will just end up turning off the rule entirely in frustration.

A fair point for sure. For this, I was going off of the text of the initial bug notice 2189 and that was what was requested, but if you've got an idea for a more productive and less noisy approach, I can make that change.

bradwilson:
[`TheoryData`](https://github.com/xunit/xunit/blob/v2/src/xunit.core/TheoryData.cs) allows strong typing when creating theory data, rather than just relying on the weakly typed `IEnumerable<object[]>`.

Example from [`ArgumentFormatterTests`](https://github.com/xunit/xunit/blob/v2/test/test.xunit.assert/Asserts/Sdk/ArgumentFormatterTests.cs):

```csharp
public static TheoryData<Type, string> TypeValueData = new()
{
    { typeof(string), "typeof(string)" },
    { typeof(int[]), "typeof(int[])" },
    { typeof(int).MakeArrayType(1), "typeof(int[*])" },
    { typeof(int).MakeArrayType(2), "typeof(int[,])" },
    { typeof(int).MakeArrayType(3), "typeof(int[,,])" },
    { typeof(DateTime[,]), "typeof(System.DateTime[,])" },
    { typeof(decimal[][,]), "typeof(decimal[][,])" },
    { typeof(IEnumerable<>), "typeof(System.Collections.Generic.IEnumerable<>)" },
    { typeof(IEnumerable<int>), "typeof(System.Collections.Generic.IEnumerable<int>)" },
    { typeof(IDictionary<,>), "typeof(System.Collections.Generic.IDictionary<,>)" },
    { typeof(IDictionary<string, DateTime>), "typeof(System.Collections.Generic.IDictionary<string, DateTime>)" },
    { typeof(IDictionary<string[,], DateTime[,][]>), "typeof(System.Collections.Generic.IDictionary<string[,], DateTime[,][]>)" },
    { typeof(bool?), "typeof(bool?)" },
    { typeof(bool?[]), "typeof(bool?[])" }
};

[Theory]
[MemberData(nameof(TypeValueData), DisableDiscoveryEnumeration = true)]
public static void TypeValue(Type type, string expected)
{
    Assert.Equal(expected, ArgumentFormatter.Format(type));
}
```

Here, the types in `TheoryData<>` must be type compatible with the arguments, and we can have a strong comparison for the nullability being compatible (i.e., mismatching `TheoryData<string?>` with a single theory argument type of `string`).

We use `TheoryData` extensively in the analyzer tests themselves. 😄 

That said, I'm pretty sure we don't have any analyzers ensuring correct usage of `TheoryData<>` at all. As I'm looking over your list:

> (xUnit1034) MemberDataAttribute references a method that returns null values, but the theory method has non-nullable parameters
> (xUnit1037) MemberDataAttribute has arguments that are null, but the corresponding parameter in the method is non-nullable
> (xUnit1035) MemberDataAttribute has arguments that are not type-compatible with the corresponding method parameter
> (xUnit1036) MemberDataAttribute has more arguments than the referenced method has parameters

I think all of xUnit1035, xUnit1036, and xUnit1037 look fine, as they analyze the parameters to the data method. The fixers for xUnit1036 and xUnit1037 look good, too.

xUnit1034 is the one that feels like it should be replaced with `TheoryData<>` related code, rather than trying to make inferences based on `IEnumerable<object[]>` vs. `IEnumerable<object?[]>` simply because the available information isn't good enough (and it's not reasonable for us to inspect all the code inside the data generator method to see whether it might ever return null, in my opinion).

I think the best strategy would be to take xUnit1035, xUnit1036, and xUnit1037 as they are now, and remove the current implementation of xUnit1034 and put it in its own PR. I'm happy to make those changes to the existing PR if you like, or you can do them.

For this PR, we can leave the "placeholder" for xUnit1034 on the assumption that it'll get filled in with the `TheoryData<>` based implementation. I suspect you could do the remaining work in just one new analyzer: ensuring the types from `TheoryData<>` are compatible with the test method parameters, which can include verifying nullability. If it's easier to split it into two analyzers, that's also acceptable. It's up to you.

Also, if you feel like you've done enough, I'll be happy to take on whatever else is left. I definitely super appreciate all the effort here, as writing and debugging analyzers and fixers is not a trivial task, which is why we get so few contributions, and you've done an excellent job of mimicking the style that I use here. ❤️

JamesTerwilliger:
@bradwilson My pleasure to help out! It was fun learning a bit more about analyzers and fixers.

One question - where does ValueTask fit in? Did you mean TheoryData?

For the moment, I'll pull 1034 and renumber accordingly, one moment.

bradwilson:
omg, yes, `TheoryData`. I just shipped 2.6.1 which was all about stripping out `ValueTask` so I've got it on my brain. 😂

JamesTerwilliger:
Clearly you `Value` the `Task` in front of you.

(crickets)

PR #166, opened 2023-11-03
[xunit2189] Final PR for adding analyzers for MemberData (TheoryData)

JamesTerwilliger:
This PR follows up on the previous two PRs, hopefully to close out https://github.com/xunit/xunit/issues/2189

The PR adds three new analyzer errors/warnings:
- xUnit1037: TheoryData has fewer type arguments than the test method has parameters (accounting for params and optional parameters)
- xUnit1038: TheoryData has more type arguments than the test method has parameters (accounting for params) 
- xUnit1039: TheoryData type argument is not compatible with the type of the corresponding test method parameter
- xUnit1040: TheoryData type argument is marked as nullable reference type, but the corresponding test method parameter is non-nullable

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,22 +31,19 @@ If you installed xUnit.net v2 2.3.0 or higher and do not wish to use the analyze
     <tr>
       <th></th>
       <th>Stable</th>
-      <th>Latest prerelease</th>
       <th>Latest CI</th>
-      <th>CI status</th>
+      <th>Build status</th>
     </tr>
   </thead>
   <tbody>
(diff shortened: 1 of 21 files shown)


JamesTerwilliger:
@bradwilson If there is anything else you can think of that is needed to close out 2189, I can probably add it to this PR if it's small.

bradwilson:
Haven't had a chance to look at the PR yet, but I suspect that we'll probably be well covered for now.

bradwilson:
Thanks!

PR #167, opened 2023-11-10
Fixes xunit/xunit#2798: xUnit2007 code fix leads to CS8920 compiler error

etherfield:
Resolves #2798

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X2000/AssertIsTypeShouldUseGenericOverloadTypeTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X2000/AssertIsTypeShouldUseGenericOverloadTypeTests.cs
@@ -1,13 +1,14 @@
-using System.Collections.Generic;
 using Microsoft.CodeAnalysis.CSharp;
-using Microsoft.CodeAnalysis.Testing;
 using Xunit;
 using Verify = CSharpVerifier<Xunit.Analyzers.AssertIsTypeShouldUseGenericOverloadType>;
 
 public class AssertIsTypeShouldUseGenericOverloadTypeTests
 {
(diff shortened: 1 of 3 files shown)


boerdereinar:
Does not work when adding extra abstraction layer:
```
public interface IInterface1 {
    public abstract void Method();
}

public interface IInterface2 : IInterface1 {
}
```

~~Also this is supported in .NET 7, but not as a preview in .NET 6.~~
Edit: Removed irrelevant info.

bradwilson:
Thanks!

I assume the language preview features used here are all coming in C# 12? If so, then I'll just wait until C# 12 is released to add this. Should only be a few ~weeks~ days...

bradwilson:
It appears this feature is not in C# 12, as it's still open: https://github.com/dotnet/csharplang/issues/4436

boerdereinar:
@bradwilson This feature has already been introduced in C# 11, not C# 12. I have no idea why the proposal is still open. See https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/static-virtual-interface-members and https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/static-abstracts-in-interfaces.md

bradwilson:
If you attempt to use it, you get told you must "enable preview features" which require attribute decoration of any affected type (or the assembly as a whole):

image

bradwilson:
Note that this is with .NET SDK 8 and C# 12.

bradwilson:
Maybe it's runtime-dependent?

boerdereinar:
The target framework in that screenshot is .NET 6. You set it to at least .NET 7 in order for it to work.

bradwilson:
Thanks!

PR #169, opened 2023-11-20
Detect when CollectionDefinition class is not in same assembly as test class (xunit/xunit#2311)

### LINKED ISSUE #2311: Analyzer: Consider detecting when CollectionDefinition class is not in same assembly as test class
I lost a couple of hours on this one today -- apparently fixtures can be in a shared assembly, but the ICollectionFixture / CollectionDefinition must be in the same assembly as the test classes decorated with it, or it won't be found by DI. This elicits a confusing message:

```
Xunit.Sdk.TestClassException
The following constructor parameters did not have matching fixture data: MyFixture fixture
```

Thank you!

edit: I eventually figured it out through stackoverflow, which pointed me to a _tiny_ note at the foot of https://xunit.net/docs/shared-context: 
 
 _Important note: Fixtures can be shared across assemblies, but collection definitions must be in the same assembly as the test that uses them._

etherfield:
Resolves [#2311](https://github.com/xunit/xunit/issues/2311)


bradwilson:
Thanks!

PR #170, opened 2023-11-23
Fix false-positive errors from xunit/xunit#2826

JamesTerwilliger:
This PR fixes a regression introduced by https://github.com/xunit/xunit/issues/2826. It tests to see if the argument to the MemberDataAttribute should be treated as a single argument rather than a params array.

@bradwilson FYI

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/MemberDataShouldReferenceValidMemberTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/MemberDataShouldReferenceValidMemberTests.cs
@@ -863,7 +863,7 @@ public class TestClass
 	}
 
 	[Fact]
-	public async void DoesNotFindWarning_WithObjectArrayArguments()
+	public async void FindsWarning_WithObjectArrayArguments()
 	{
 		var source = $@"
 using System.Collections.Generic;
@@ -883,7 +883,21 @@ public class TestClass
 }}
 ";
 
-		await Verify.VerifyAnalyzer(LanguageVersion.CSharp10, source);
+		DiagnosticResult[] expected =
+		{
+			Verify
+				.Diagnostic("xUnit1035")
+				.WithSpan(11, 52, 11, 53)
+				.WithArguments("seq", "System.Collections.Generic.IEnumerable<object>")
+				.WithSeverity(DiagnosticSeverity.Error),
+			Verify
+				.Diagnostic("xUnit1036")
+				.WithSpan(11, 55, 11, 56)
+				.WithArguments("2")
+				.WithSeverity(DiagnosticSeverity.Error)
+		};
+
+		await Verify.VerifyAnalyzer(LanguageVersion.CSharp10, source, expected);
 	}
 
 	[Theory]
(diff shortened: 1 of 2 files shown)


src/xunit.analyzers.tests/Analyzers/X1000/MemberDataShouldReferenceValidMemberTests.cs
```diff
@@ -837,6 +837,55 @@ public class TestClass {{
...
+using System.Collections.Generic;
+using Xunit;
+
+public class TestClass
+{{
+  public static IEnumerable<object[]> GetSequences(IEnumerable<object> seq) =>
+    new List<object[]>();
+
+  [Theory]
+  [MemberData(nameof(GetSequences), new object[] {{ 1, 2 }})]
```
viceroypenguin, line 878: Technically this case _should_ have a warning. The `int[]` should not have warning, because it becomes the first parameter passed to a `params object[] parameters`. However, an `object[]` binds tighter than the `params`, so this will actually pass the parameters `1` and `2` to the method instead of `object[] { 1, 2 }`. 

NB: Confirmed via testing - top method here works successfully, but with the xUnit warnings; but this version fails with an `ArgumentException`.

JamesTerwilliger: @viceroypenguin Gotcha, will try another approach

JamesTerwilliger: @viceroypenguin This one should work

viceroypenguin (approved):
Tests look good as defined.

src/xunit.analyzers/X1000/MemberDataShouldReferenceValidMember.cs
```diff
@@ -234,7 +239,12 @@ public class MemberDataShouldReferenceValidMember : XunitDiagnosticAnalyzer
 		if (initializer is null)
 			return new List<ExpressionSyntax> { argumentExpression };
 
-		return initializer.Expressions.ToList();
+		// In the special case where the argument is an object[], treat like params
+		var type = semanticModel.GetTypeInfo(argumentExpression).Type;
+		if (type is IArrayTypeSymbol arrayType && arrayType.ElementType.SpecialType == SpecialType.System_Object)
+			return initializer.Expressions.ToList();
+
+		return new List<ExpressionSyntax> { argumentExpression };
```
viceroypenguin, line 247: @JamesTerwilliger could this be shortcutted by inverting the `if` on 244 and executing the `if ` between line 231 and 232? then in this case, it does not have to determine `initializer`. Thoughts?

JamesTerwilliger: @viceroypenguin I had it this way thinking that initializer being not null means that it is an array type, and that the semantic model call is likely more expensive. Said differently, the initializer being null is the short-circuit that I want so that the (I believe) expensive call is only done if need be.

viceroypenguin: Good reasoning; and I'll trust your experience with this code, not having enough experience with touching the `semanticModel` in analyzers. LGTM

bradwilson:
Thanks!

PR #172, opened 2023-11-29
Verify fixtures have sources (and mention same-assembly requirement for xunit/xunit#2311)

### LINKED ISSUE #2311: Analyzer: Consider detecting when CollectionDefinition class is not in same assembly as test class

etherfield:
Resolves [2311](https://github.com/xunit/xunit/issues/2311)


bradwilson:
@etherfield Looking over this source and understanding what's going on here, it occurs to me what's really being done here: verifying that fixtures can be provided at runtime. The need for the collection definition to be in the same assembly is a related but orthogonal requirement which can be surfaced during this discovery. It occurred to me as I realized that this code needed to be updated for `IClassFixture<>` support what it was really doing.

Since this is about 80% of the way there, I'm going to make the pivot and adapt it to do the primary task of ensuring fixture sources are specified with the secondary task to letting people know when their fixture sources can't be found that it may be caused by the fact that their collection definition lives in another assembly.

PR #173, opened 2023-12-01
Adding X2024 and X2025 analyzers for Assert.True and Assert.False

JamesTerwilliger:
@bradwilson This PR addresses, I believe, https://github.com/xunit/xunit/issues/1503

It adds a new analyzer and associated fixers to test for cases that look like:
`Assert.True(expression == literal)`
`Assert.True(expression != null)`
`Assert.True(expression == MyEnum.Value)`

and suggests to turn them into (and fixers actually turn them into):
`Assert.Equal(literal, expression)`
`Assert.NotNull(expression)`
`Assert.Equal(MyEnum.Value, expression)`

Some notes about usage:
- Because there is no option for "message" in Equals, Null, etc. we don't raise the issue if there is a message.
- Setting these to Info for the moment, but open to having them be Warning
- Should detect whether the literal is on the left or the right, and when generating the Equal/NotEqual, always put the literal on the left
- Only works for literal expressions (char, decimal, double, float, string, bool, etc.) and direct references to Enum values


bradwilson:
I think I'm okay with these being `Info` for now.

bradwilson:
Thanks!

PR #174, opened 2023-12-04
Updates Microsoft.CodeAnalysis.CSharp version to latest

JamesTerwilliger:
This PR resolves issue https://github.com/xunit/xunit/issues/2789 by updating the NuGet version of the code analysis tools to the latest, so that CSharp12 is included. The new version of the NuGet package supports Framework4.7.2 and .NetStandard2.0 so should not introduce any compatibility issues.

One small code change was required as using CultureInfo.CurrentCulture in the way used in that fixer is now listed as a "banned symbol" because the compiler settings will pick up the culture. Feel free to revert that part of the change to see the error if you'd like @bradwilson :-).

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -23,7 +23,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="[4.2.0]" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="[4.8.0]" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="all" />
     <PackageReference Include="Nerdbank.GitVersioning" Version="3.6.133">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/src/xunit.analyzers.fixes/X2000/BooleanAssertsShouldNotBeNegatedFixer.cs
+++ b/src/xunit.analyzers.fixes/X2000/BooleanAssertsShouldNotBeNegatedFixer.cs
@@ -41,7 +41,7 @@ public class BooleanAssertsShouldNotBeNegatedFixer : BatchedCodeFixProvider
 
 		context.RegisterCodeFix(
 			CodeAction.Create(
-				string.Format(CultureInfo.CurrentCulture, "Use Assert.{0}", replacement),
+				string.Format("Use Assert.{0}", replacement),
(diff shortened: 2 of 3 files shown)


bradwilson:
Unfortunately this would mean the analyzers would only be supported on VS 2022 17.8 or later: https://learn.microsoft.com/en-us/visualstudio/extensibility/roslyn-version-support?view=vs-2022

The current version (4.2.0) was selected because VS 2022 17.2 is the currently oldest still supported LTS version, until January 2024 at least: https://learn.microsoft.com/en-us/visualstudio/productinfo/vs-servicing

bradwilson:
The removal of `CultureInfo.CurrentCulture` definitely feels like an error. Can you link to something that explains why you wanted to remove it?

JamesTerwilliger:
> The removal of `CultureInfo.CurrentCulture` definitely feels like an error. Can you link to something that explains why you wanted to remove it?

@bradwilson Well, "wanted" is not exactly accurate, as noted in the PR description, without removing it, compiling fails with an error:

`Severity	Code	Description	Project	File	Line	Suppression State	Details
Error	RS1035	The symbol 'CultureInfo.CurrentCulture' is banned for use by analyzers: Analyzers should use the locale given by the compiler command line arguments, not the CurrentCulture	xunit.analyzers.fixes	C:\repos\xunit.analyzers\src\xunit.analyzers.fixes\X2000\BooleanAssertsShouldNotBeNegatedFixer.cs	44	Active	
`
This banned symbol must be new to 4.8.0 or at least post-4.2.0.

An alternative question may be, "why does `BooleanAssertsShouldNotBeNegatedFixer` use CultureInfo.CurrentCulture but none of the other fixers do"?

As this PR won't go in due to the VS support issue, it's sort of a moot point, but it does seem that either every fixer should use CultureInfo (and in doing so hit the banned symbol buzzsaw issue once it can update to 4.8.0), or none of them should, or they should all use some other future-proofed approach to do the same thing.

bradwilson:
> An alternative question may be, "why does BooleanAssertsShouldNotBeNegatedFixer use CultureInfo.CurrentCulture but none of the other fixers do"?

That's mostly because the analyzers haven't had the full .NET analyzer suite enabled (yet). It's a habit that I've gotten into since I did so with the v3 core source (and also removing all instances of string interpolation). [CA13xx rules](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/globalization-warnings) basically force the use of cultures and relation options (like string comparison choices).

JamesTerwilliger:
Interesting. It does make me wonder what to do when the code does move to 4.8.0 eventually and that pattern is considered banned, but as noted, that's now a problem for future us.

PR #175, opened 2023-12-05
Create analyzer for constructors of types derived from FactAttribute

JamesTerwilliger:
@bradwilson Proposed analyzer and fixer for issue https://github.com/xunit/xunit/issues/2120

Looks for constructors on classes that inherit from FactAttribute and makes sure that there are no internal or protected internal constructors. I make the assumption that if for some reason there are private or protected constructors that those are there for a reason.

Certainly open to other logic as to what visibility modifiers are "OK" here, as well as whether it's better to mark the class itself or the usage of the class inside an attribute.


bradwilson:
I think the logic probably needs to go like this:

Inspect the Fact (or Fact-derived) attribute and discover what constructor the user is calling, and if that constructor is not public, then raise the analyzer.

The mere presence of an internal constructor is not necessarily a problem, unless the developer tries to use it for their test. It's also possible (likely, in the case of Fact and Theory) that the Fact-derived attribute doesn't have any constructors, so that scenario also needs to be supported.

bradwilson:
Thoughts about adapting this? Are you still interested in doing this @JamesTerwilliger?

JamesTerwilliger:
> Thoughts about adapting this? Are you still interested in doing this @JamesTerwilliger?

Apologies, I've been underwater the past few weeks. I can still take this on. With the suggested approach, we can certainly flag the uses of the constructor, but I'm not sure we'd also be able to have a fixer. I'll at least get the analyzer in place without a fixer and start there.

bradwilson:
I don't think a fixer is necessary.

JamesTerwilliger:
> I don't think a fixer is necessary.

@bradwilson In that case, I think this is ready for review now

bradwilson:
Thanks!

PR #176, opened 2023-12-15
Analyzer for equality checks with hash sets

### LINKED ISSUE #2840: Add an analyzer for equality checks with hash sets + a comparer func
Now that https://github.com/xunit/xunit/issues/2838 has been resolved, we should write an analyzer which flags as an error any time the user writes an equality comparison using either of these two functions:

```csharp
void Equal<T>(IEnumerable<T> expected, IEnumerable<T> actual, Func<T, T, bool> comparer)
void NotEqual<T>(IEnumerable<T> expected, IEnumerable<T> actual, Func<T, T, bool> comparer)
```

when `expected` and `actual` are know to be sets (that is, something which implements either `ISet<>` or `IReadOnlySet<>`).

etherfield:
Resolves [#2840](https://github.com/xunit/xunit/issues/2840)

The most basic case is covered: built-in hashsets. Covering user-defined types would be quite a hassle.
Given that, I presume it's better off having the analyzer throw a warning rather than an error.

edit: by covering user-defined types I meant checking if GetHashCode() is overridden correctly, which is out of scope anyway. So nevermind, it's covered by checking ISet and IReadOnlySet.


bradwilson:
Thanks for this!

I'm going to merge main (to pick up the new multi-targeting for Roslyn support) and update the tests to use matrix theory data to simplify them a bit. At the same time, I'm also going to use explicit types in the analyzer rather than just comparing names against literal `ISet` and `IReadOnlySet`.

bradwilson:
I also added tests (and updated the code) to handle the case where the collection classes in question weren't generic (i.e., `public class MyDataStructure : ISet<int>`).

bradwilson:
Thanks!

PR #177, opened 2024-01-02
xUnit1030 should not trigger inside lambda

MartyIX:
In my project, I was forced to turn off the xUnit1030 analyzer because it does not take into account scenarios like the one in the test case. I figured that a PR is the easiest way to demonstrate my issue.

However, I'm not entirely sure if the `xUnit1030` analyzer is wrong here, or if I'm wrong and the test sample is incorrect. 

Could you comment on it @bradwilson, please?


bradwilson:
Looks like the same kind of bug as xunit/xunit#2856.

bradwilson:
Available in `1.9.0-pre.10` https://xunit.net/docs/using-ci-builds

PR #178, opened 2024-01-06
xUnit1030/xUnit1031 should not trigger inside local functions

MartyIX:
Follow-up to #177


bradwilson:
Thanks, I'm going to piggyback the same fix for xUnit1031 as well, since they both got the lambda fix.

bradwilson:
Available in `1.10.0-pre.5` https://xunit.net/docs/using-ci-builds

PR #179, opened 2024-01-26
Fix for https://github.com/xunit/xunit/issues/2873

JamesTerwilliger:
IsAssignableFrom logic does not account for the possibility that tuple type fields may be named differently. This change augments the logic to allow for the names to not be equal.


bradwilson:
Thanks!

PR #180, opened 2024-02-07
Fix analyzer to account for issue 2557

### LINKED ISSUE #2886: Update xUnit1041 to support generic fixtures
Using the generic collection fixture feature enabled by #2557 currently triggers xUnit1041.

```csharp
using Xunit;

public class GenericClass<T1, T2>
{ }

[CollectionDefinition("GenericFixture")]
public class GenericFixture<T1, T2> : ICollectionFixture<GenericClass<T1, T2>>
{ }

[Collection("GenericFixture")]
public class TestClass
{
    public TestClass(GenericClass<string, int> fixture)
    { }

    [Fact]
    public void TestMethod()
    { }
}
```

image

JamesTerwilliger:
With https://github.com/xunit/xunit/issues/2557 being closed as done, the analyzer now also needs to be updated to account for this use case. This PR edits the logic for xUnit1041 to allow for the generic case outlined in that issue.

Fixes https://github.com/xunit/xunit/issues/2886

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/EnsureFixturesHaveASourceTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/EnsureFixturesHaveASourceTests.cs
@@ -186,13 +186,35 @@ public class Fixture { }
 [CollectionDefinition(""test"")]
 public class TestCollection : ICollectionFixture<Fixture> { }
 
-[Collection(""test"")]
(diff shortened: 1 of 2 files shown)


bradwilson:
That was quick! Thanks! 😄 

I'm adding a second test (the simpler version), and then making it so the new code only applies when you're using xUnit.net v3 (since we don't support this in v2).

PR #183, opened 2024-03-07
Create analyzer for non-serializable TheoryData type arguments (xunit/xunit#2866)

jared-chevalier:
Suggested initial version of a solution for issue [xunit/xunit#2866](https://github.com/xunit/xunit/issues/2866).

Includes analyzer for proposed new rule xUnit1044: TheoryData type arguments should be serializable.

Please let me know if you have any feedback.

# Solution

## `Serializability` enumeration

- `NeverSerializable`
- `PossiblySerializable`
- `AlwaysSerializable`

## `TheoryDataTypeArgumentFinder` class

- Finds all `TheoryData<>` type arguments for a data source referenced by a given `[ClassData]` or `[MemberData]` attribute in a given test class, if applicable
- If the data source's type is not compatible with a `TheoryData<>` class (such as if it is a member with the type `IEnumerable<object[]>`), then no type arguments will be found
- If the data source's type derives from a `TheoryData<>` base class, then the type arguments for the base class will be found 

## `SerializabilityAnalyzer` class

- Analyzes a given type to determine whether it is always, possibly, or never serializable
- The static analysis of a type's serializability closely corresponds to the logic found in `SerializationHelper` and `MemberDataAttributeBase` from the `xunit` repository
- See discussion in issue [xunit/xunit#2866](https://github.com/xunit/xunit/issues/2866)

## `TypeSymbols` class

- Lazy caching wrapper around `TypeSymbolFactory` (plus the `Compilation` and `XunitContext`), for reusing already generated type symbols

## `TheoryDataTypeArgumentsShouldBeSerializable` analyzer

- Analyzes `MethodDeclarationSyntax` nodes with a `[Theory]` attribute
- For each `[ClassData]` or `[MemberData]` attribute: Uses `TheoryDataTypeArgumentFinder` to find all `TheoryData<>` type arguments for the referenced class or member, if applicable
- For each type argument found: Uses `SerializabilityAnalyzer` to check whether the type is serializable (unless the type should be ignored)
- For each never serializable type argument: Reports an xUnit1044 diagnostic for the data attribute syntax node, with a strongly phrased message: `The type argument {0} is not serializable. Consider using a type that is known to be serializable.`
- For each possibly serializable type argument: Reports an xUnit1044 diagnostic for the data attribute syntax node, with a weakly phrased message: `The type argument {0} might not be serializable. Consider using a type that is known to be serializable.`

# Key points

## Diagnostic severity

- The diagnostic currently has `Info` severity. Would `Warning` be more appropriate?
- Currently, using `Warning` would break the build because an existing test class, `SetEqualityAnalyzerTests`, uses a tuple type argument for a `MatrixTheoryData` data source, and warnings are treated as errors
- A solution would be to skip analyzing data sources associated with `DisableDiscoveryEnumeration = true`, but that has not been implemented yet (see discussion in issue [xunit/xunit#2866](https://github.com/xunit/xunit/issues/2866))

## Statically analyzing serializability of enumerations

- See discussion in issue [xunit/xunit#2866](https://github.com/xunit/xunit/issues/2866)
- Any given enumeration type is either always serializable (from a local assembly) or never serializable (from the GAC)
- As far as I can tell, static analysis cannot determine whether an enumeration type is from a local assembly or from the GAC
- As a result, all enumeration types are treated as possibly serializable
- However, the analyzer ignores them, in order to prevent a diagnostic from being found for every single enumeration type

## IXunitSerializable type symbol

- `IXunitSerializable` is in namespace `Xunit.Abstractions` in xUnit version 2, but in `Xunit.Sdk` in version 3
- The version 2 type symbol is accessible via a lazy cache in `xunitContext.V2Abstractions?.IXunitSerializableType` (as well as directly via `TypeSymbolFactory`)
- There does not seem to be any corresponding property in any of the version 3 contexts. Would it make sense to add one somewhere in the contexts? Ideally there would be a common interface, implemented for each version, that provided the type symbol for the current version
- I added the `TypeSymbolFactory.IXunitSerializable_V3` factory method for the version 3 type symbol
- My `TypeSymbols` class currently uses the following logic to access the appropriate type symbol: `TypeSymbolFactory.IXunitSerializable_V3(compilation) ?? xunitContext.V2Abstractions?.IXunitSerializableType`

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializableTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializableTests.cs
@@ -350,7 +350,7 @@ public class PossiblySerializableUnsealedClass {{ }}";
 		}
 	}
 
-	public sealed class X1044_TheoryDataTypeArgumentsShouldBeSerializable : TheoryDataTypeArgumentsShouldBeSerializableTests
+	public sealed class X1044_AvoidUsingTheoryDataTypeArgumentsThatAreNotSerializable : TheoryDataTypeArgumentsShouldBeSerializableTests
 	{
 		const string Id = "xUnit1044";
 
@@ -431,7 +431,7 @@ public struct NonSerializableStruct {{ }}";
 		}
 	}
 
-	public sealed class X1045_TheoryDataTypeArgumentsShouldBeDefinitelySerializable : TheoryDataTypeArgumentsShouldBeSerializableTests
+	public sealed class X1045_AvoidUsingTheoryDataTypeArgumentsThatMightNotBeSerializable : TheoryDataTypeArgumentsShouldBeSerializableTests
 	{
 		const string Id = "xUnit1045";
 
--- a/src/xunit.analyzers.tests/Analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializableTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializableTests.cs
@@ -145,6 +145,20 @@ public class DataSource<T1, T2> : IEnumerable<object[]> {
 	[MemberData(nameof(TheoryDataMembers), "SerializableEnumeration")]
 	[MemberData(nameof(TheoryDataMembers), "SerializableEnumeration?")]
 	[MemberData(nameof(TheoryDataMembers), "Dictionary<string, List<string>>")]
+
+#if NET6_0_OR_GREATER && ROSLYN_4_4_OR_GREATER
+
+	[MemberData(nameof(TheoryDataMembers), "DateOnly")]
+	[MemberData(nameof(TheoryDataMembers), "DateOnly[]")]
+	[MemberData(nameof(TheoryDataMembers), "DateOnly?")]
+	[MemberData(nameof(TheoryDataMembers), "DateOnly?[]")]
+	[MemberData(nameof(TheoryDataMembers), "TimeOnly")]
+	[MemberData(nameof(TheoryDataMembers), "TimeOnly[]")]
+	[MemberData(nameof(TheoryDataMembers), "TimeOnly?")]
+	[MemberData(nameof(TheoryDataMembers), "TimeOnly?[]")]
+
+#endif
+
 	public async void GivenTheoryMethod_WithSerializableTheoryDataMember_DoesNotFindDiagnostic(
 		string member,
 		string attribute,
@@ -218,39 +232,6 @@ public struct SerializableStruct : ISerializableInterface {{
 }}";
 		}
 	}
-
-#if NET6_0_OR_GREATER && ROSLYN_4_4_OR_GREATER
-
-	[Theory]
-	[MemberData(nameof(TheoryDataMembers), "DateOnly")]
-	[MemberData(nameof(TheoryDataMembers), "DateOnly[]")]
-	[MemberData(nameof(TheoryDataMembers), "DateOnly?")]
-	[MemberData(nameof(TheoryDataMembers), "DateOnly?[]")]
-	[MemberData(nameof(TheoryDataMembers), "TimeOnly")]
-	[MemberData(nameof(TheoryDataMembers), "TimeOnly[]")]
-	[MemberData(nameof(TheoryDataMembers), "TimeOnly?")]
-	[MemberData(nameof(TheoryDataMembers), "TimeOnly?[]")]
-	public async void GivenTheoryMethod_WithDateOnlyOrTimeOnlyInTheoryDataMember_DoesNotFindDiagnostic(
-		string member,
-		string attribute,
-		string type)
-	{
-		var source = $@"
-using System;
-using Xunit;
-
-public class TestClass {{
-    {member}
-
-    [Theory]
-    [{attribute}]
-    public void TestMethod({type} parameter) {{ }}
-}}";
-
-		await Verify.VerifyAnalyzer(LanguageVersion.CSharp10, source);
-	}
-
-#endif
 
 	[Theory]
 	[MemberData(nameof(TheoryDataMembers), "object")]
--- a/src/xunit.analyzers.tests/Analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializableTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializableTests.cs
@@ -1,11 +1,10 @@
-using Microsoft.CodeAnalysis.CSharp;
 using Xunit;
 using Verify = CSharpVerifier<Xunit.Analyzers.TheoryDataTypeArgumentsShouldBeSerializable>;
 
 public class TheoryDataTypeArgumentsShouldBeSerializableTests
 {
-	const string IsNotSerializable = "is not serializable";
-	const string MightNotBeSerializable = "might not be serializable";
+	const string X1044 = "xUnit1044";
+	const string X1045 = "xUnit1045";
 
 	public static TheoryData<string, string, string, string> TheoryDataClass(
 		string type1,
@@ -234,6 +233,48 @@ public struct SerializableStruct : ISerializableInterface {{
 	}
 
 	[Theory]
+	[MemberData(nameof(TheoryDataMembers), "Delegate")]
+	[MemberData(nameof(TheoryDataMembers), "Delegate[]")]
+	[MemberData(nameof(TheoryDataMembers), "Func<int>")]
+	[MemberData(nameof(TheoryDataMembers), "Func<int>[]")]
+	[MemberData(nameof(TheoryDataMembers), "NonSerializableSealedClass")]
+	[MemberData(nameof(TheoryDataMembers), "NonSerializableSealedClass[]")]
+	[MemberData(nameof(TheoryDataMembers), "NonSerializableStruct")]
+	[MemberData(nameof(TheoryDataMembers), "NonSerializableStruct[]")]
+	[MemberData(nameof(TheoryDataMembers), "NonSerializableStruct?")]
+	[MemberData(nameof(TheoryDataMembers), "NonSerializableStruct?[]")]
+	public async void GivenTheoryMethod_WithNonSerializableTheoryDataMember_FindsDiagnosticX1044(
+		string member,
+		string attribute,
+		string type)
+	{
+		var source = $@"
+using System;
+using System.Text;
+using Xunit;
+
+public class TestClass {{
+    {member}
+
+    [Theory]
+    [{attribute}]
+    public void TestMethod({type} parameter) {{ }}
+}}
+
+public sealed class NonSerializableSealedClass {{ }}
+
+public struct NonSerializableStruct {{ }}";
+
+		var expected =
+			Verify
+				.Diagnostic(X1044)
+				.WithSpan(10, 6, 10, 6 + attribute.Length)
+				.WithArguments(type);
+
+		await Verify.VerifyAnalyzer(source, expected);
+	}
+
+	[Theory]
 	[MemberData(nameof(TheoryDataMembers), "object")]
 	[MemberData(nameof(TheoryDataMembers), "object[]")]
 	[MemberData(nameof(TheoryDataMembers), "Array")]
@@ -250,7 +291,7 @@ public struct SerializableStruct : ISerializableInterface {{
 	[MemberData(nameof(TheoryDataMembers), "IPossiblySerializableInterface[]")]
 	[MemberData(nameof(TheoryDataMembers), "PossiblySerializableUnsealedClass")]
 	[MemberData(nameof(TheoryDataMembers), "PossiblySerializableUnsealedClass[]")]
-	public async void GivenTheoryMethod_WithPossiblySerializableTheoryDataMember_FindsWeakDiagnostic(
+	public async void GivenTheoryMethod_WithPossiblySerializableTheoryDataMember_FindsDiagnosticX1045(
 		string member,
 		string attribute,
 		string type)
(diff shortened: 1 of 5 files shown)


src/xunit.analyzers.tests/Analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializableTests.cs
```diff
@@ -0,0 +1,426 @@
...
+	[Theory]
+	[MemberData(nameof(TheoryDataMembers), "DateOnly")]
+	[MemberData(nameof(TheoryDataMembers), "DateOnly[]")]
+	[MemberData(nameof(TheoryDataMembers), "DateOnly?")]
+	[MemberData(nameof(TheoryDataMembers), "DateOnly?[]")]
+	[MemberData(nameof(TheoryDataMembers), "TimeOnly")]
+	[MemberData(nameof(TheoryDataMembers), "TimeOnly[]")]
+	[MemberData(nameof(TheoryDataMembers), "TimeOnly?")]
+	[MemberData(nameof(TheoryDataMembers), "TimeOnly?[]")]
+	public async void GivenTheoryMethod_WithDateOnlyOrTimeOnlyInTheoryDataMember_DoesNotFindDiagnostic(
```
bradwilson: You could collapse this into `GivenTheoryMethod_WithSerializableTheoryDataMember_DoesNotFindDiagnostic` and just put the member data itself inside the `#if` block.

src/xunit.analyzers/Utility/Descriptors.xUnit1xxx.cs
```diff
@@ -402,7 +402,14 @@ public static partial class Descriptors
 			"Constructor '{0}' must be public to be used on a test method."
 		);
 
-	// Placeholder for rule X1044
+	public static DiagnosticDescriptor X1044_TheoryDataTypeArgumentsShouldBeSerializable { get; } =
```
bradwilson: As discussed in the issue thread, I believe it would be best to emit two different diagnostics here, so users can turn the "maybe" one off without having to turn the "definitely" one off.

src/xunit.analyzers/Utility/Descriptors.xUnit1xxx.cs
```diff
@@ -402,7 +402,14 @@ public static partial class Descriptors
...
+		Diagnostic(
+			"xUnit1044",
+			"TheoryData type arguments should be serializable",
+			Usage,
+			Info,
+			"The type argument {0} {1}. Consider using a type that is known to be serializable."
```
bradwilson: The two diagnostics should get the updated language from the issue thread.

src/xunit.analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializable.cs
```diff
@@ -0,0 +1,567 @@
...
+		base(Descriptors.X1044_TheoryDataTypeArgumentsShouldBeSerializable)
+	{ }
+
+	public override void AnalyzeCompilation(
+		CompilationStartAnalysisContext context,
+		XunitContext xunitContext)
+	{
+		Guard.ArgumentNotNull(context);
+		Guard.ArgumentNotNull(xunitContext);
+
```
bradwilson, line 28: We can skip early for any project which isn't targeting .NET Framework.

src/xunit.analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializable.cs
```diff
@@ -0,0 +1,567 @@
...
+			}
+
+			return null;
+		}
+
+		INamedTypeSymbol? GetDataSourceType(
+			AttributeData dataAttribute,
+			INamedTypeSymbol testClass)
+		{
+			if (dataAttribute.IsInstanceOf(typeSymbols.ClassDataAttribute))
```
bradwilson: This should be an exact type match, since any class derived from `ClassDataAttribute` does not necessarily have to have a constructor of this shape.

src/xunit.analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializable.cs
```diff
@@ -0,0 +1,567 @@
...
+				return dataAttribute.ConstructorArguments.FirstOrDefault().Value as INamedTypeSymbol;
+
+			if (dataAttribute.IsInstanceOf(typeSymbols.MemberDataAttribute))
```
bradwilson: Even though `MemberDataAttribute` is currently sealed, philosophically I'd still like to see an exact match here as well. If it turns out we decide to unseal it for v3, I'd want this code to continue to behave in the correct way; this is not currently planned, but I'd rather be safe. 😄

bradwilson (changes requested):
Looking very good. Just a few comments based on the thread in the associated issue.

src/xunit.analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializable.cs
jared-chevalier: Can you explain this? I don't understand the reasoning here.

src/xunit.analyzers.tests/Analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializableTests.cs
jared-chevalier: Good point.

src/xunit.analyzers/Utility/Descriptors.xUnit1xxx.cs
jared-chevalier: Sounds good.

src/xunit.analyzers/Utility/Descriptors.xUnit1xxx.cs
jared-chevalier: Sounds good.

src/xunit.analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializable.cs
jared-chevalier: Good point!

src/xunit.analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializable.cs
jared-chevalier: No problem, makes sense!

src/xunit.analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializable.cs
bradwilson: The GAC only exists in .NET Framework. It doesn't exist for .NET Core/.NET.

bradwilson: I realize now I really put this comment in the wrong place and with nowhere near enough words. 😝 We can skip "Enum in the GAC" analysis when the context is analyzing a .NET Framework targeting project.

jared-chevalier: Ok, that makes more sense! 😂

jared-chevalier: Could you clarify how to determine whether the context is analyzing a project targeting .NET Framework?

bradwilson: I've never done it myself, so I don't know. I assume it would be part of the context that's given, but I'd have to poke around it in a debugger to figure out how (and whether it's even possible), which is how I do 100% of my Roslyn work, sadly. 😂

bradwilson: Since you don't know, we can leave it off for now and I'll try to add it later.

jared-chevalier: > which is how I do 100% of my Roslyn work, sadly.

😂

I poked around the context aggregate, and briefly searched online, but didn't find anything.

I'll spend a little time later investigating how feasible it would be to check whether Roslyn has source-level access to an enum definition, like you suggested.

If it's not, and if we can't figure out how to check the target framework, then I suppose the enum serializability problem is intractable after all. In which case, the analyzer should presumably continue ignoring all enum types? Since otherwise, every single enum type would produce the "might not be serializable" diagnostic, which I imagine would be more bothersome than useful.

src/xunit.analyzers/Utility/Descriptors.xUnit1xxx.cs
```diff
@@ -402,9 +402,23 @@ public static partial class Descriptors
...
 
-	// Placeholder for rule X1045
+	public static DiagnosticDescriptor X1045_TheoryDataTypeArgumentsShouldBeDefinitelySerializable { get; } =
+		Diagnostic(
+			"xUnit1045",
+			"TheoryData type arguments should be definitely serializable",
+			Usage,
+			Info,
+			"The type argument {0} might not be serializable, which may cause Test Explorer to not enumerate individual data rows. Consider using a type that is known to be serializable."
+		);
```
jared-chevalier: Do you have any alternative suggestions for titles and property names for these two diagnostics?

I tried to come up with a pair of titles that clearly distinguished between the two without excessive wordiness or double negatives. This is what I came up with after trying out some options, but I'm not sure I'm happy with it.

src/xunit.analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializable.cs
bradwilson: > I'll spend a little time later investigating how feasible it would be to check whether Roslyn has source-level access to an enum definition, like you suggested.

I was able to dig this out in the debugger. 😁 Looks like you can look for `ISourceAssemblySymbol` on the type's `ContainingAssembly`. Here's one where the type is local:

```
> type.ContainingAssembly is ISourceAssemblySymbol
true
```

And here's one where it isn't:

```
> context.Compilation.GetTypeByMetadataName("System.ConsoleColor").ContainingAssembly is ISourceAssemblySymbol
false
```

The actual concrete types were helpfully named `SourceAssemblySymbol` and `NonSourceAssemblySymbol` 😆 

Still looking for access to target framework...

bradwilson: I think the more I think about it, we're only circling around to something vaguely successful for enums, so we should just ignore enums entirely. The GAC requirement isn't reliably testable, and honestly most people these days are probably targeting .NET (and not .NET Framework), so let's just let the .NET Framework people fend for themselves, which is exactly where they would've been before this analyzer.

src/xunit.analyzers/Utility/Descriptors.xUnit1xxx.cs
bradwilson: Yeah, this one's weird. It's almost like it's better to invert the naming to be the problem rather than the solution because "Should" is applies to both.

Maybe:

`X1044_TheoryDataTypeArgumentMightNotBeSerializable`
`X1045_TheoryDataTypeArgumentIsNotSerializable`

With titles:

`TheoryData type arguments might not be serializable`
`TheoryData type arguments are not serializable`

Thoughts?

src/xunit.analyzers/X1000/TheoryDataTypeArgumentsShouldBeSerializable.cs
jared-chevalier: > The actual concrete types were helpfully named `SourceAssemblySymbol` and `NonSourceAssemblySymbol` 😆

Thanks, good to know!

> I think the more I think about it, we're only circling around to something vaguely successful for enums, so we should just ignore enums entirely. The GAC requirement isn't reliably testable, and honestly most people these days are probably targeting .NET (and not .NET Framework), so let's just let the .NET Framework people fend for themselves, which is exactly where they would've been before this analyzer.

Alright, sounds good.

src/xunit.analyzers/Utility/Descriptors.xUnit1xxx.cs
jared-chevalier: I went through the existing diagnostics and did an inventory of the language used, to see whether there is a precedent for this.

Most use standard prescriptive language: `must (not)`, `should (not)`, `can(not)`, `is recommended`.

A few use imperative voice: `ensure`, `do not`, `use`.

Only one uses descriptive language in the property name, describing the problem rather than prescribing the solution: `X1028_TestMethodHasInvalidReturnType` (though the title is phrased prescriptively: `Test method must have valid return type`).

I like your suggestions, as they are clear and simple.

Here's another option, using imperative voice with `avoid using`:

`Avoid using TheoryData type arguments that are not serializable`
`Avoid using TheoryData type arguments that might not be serializable`

I suppose mine are slightly more consistent with existing style, but also slightly more wordy. I'm happy to go with whichever you prefer.

jared-chevalier: By the way, did you intend to switch the diagnostic ids? I currently have X1044 as the "is not serializable" diagnostic and X1045 as the "might not be serializable" one.

bradwilson: No, that was unintentional.

bradwilson: I like your suggestions. 👍🏼

jared-chevalier: Ok, sounds good.

bradwilson (approved):
Awesome job! Thanks

bradwilson:
I decided to take this work and make parallel analyzers for v3's `TheoryDataRow` constructor: https://github.com/xunit/xunit.analyzers/commit/68525f89d851fb482221f3ffdf0560929a9be30d

jared-chevalier:
> Awesome job! Thanks

You're welcome, it was my pleasure!

> I decided to take this work and make parallel analyzers for v3's `TheoryDataRow` constructor: [68525f8](https://github.com/xunit/xunit.analyzers/commit/68525f89d851fb482221f3ffdf0560929a9be30d)

Nice!

PR #185, opened 2024-07-21
Create analyzer/fixer for Assert.NotEmpty

RieBi:
So this is intended to solve the other part of xunit/xunit#1510, but I've got a few questions to @bradwilson before doing it.

First of all, I didn't realize when I was doing first half that they are almost the same, so I wonder how best would be to do it. I assume it would be fine to change the existing classes to account for both descriptors, but want to make sure if that would be fine.

As a side thing, [this method name](https://github.com/xunit/xunit.analyzers/blob/729501d68c9b3b932ea171828131cb7e45bb9395/src/xunit.analyzers.tests/Analyzers/X2000/AssertEmptyShouldNotBeUsedForCollectionDoesNotContainCheckTests.cs#L100) assumes that the analyzer does not trigger on strings, while in fact it does trigger. So, is that just a name thing and in fact it should trigger, or it is not intended to be triggered on strings and should be changed as well?

These are the things I want to clarify, and afterwards I think I'll work on finishing resolving the issue.


bradwilson:
> I assume it would be fine to change the existing classes to account for both descriptors, but want to make sure if that would be fine.

Absolutely!

And apologies if the code looks different; I finally got motivated to standardize the many years worth of code style drift, motivated by discovering `/* lang=c#-test */` syntax highlighting. 😄 

> As a side thing, [this method name](https://github.com/xunit/xunit.analyzers/blob/729501d68c9b3b932ea171828131cb7e45bb9395/src/xunit.analyzers.tests/Analyzers/X2000/AssertEmptyShouldNotBeUsedForCollectionDoesNotContainCheckTests.cs#L100) assumes that the analyzer does not trigger on strings, while in fact it does trigger.

You are correct! In my attempt to fix up all the naming (and I'll note sheepishly I missed the test above it), I put the wrong text there. It is intended to trigger. I have just pushed up name changes for both methods. 🙇🏼‍♂️

RieBi:
This is it, I guess.

If the tests are correct, which they should be, it finds the following situation:

`Assert.NotEmpty(Enumerable.Where(expression))`

and suggest to change to:

`Assert.Contains(Enumerable, expression)`

I'm looking forward to receive the feedback on it, if there are any issues with my implementation.

bradwilson:
Implementation looks great! My only changes are for me being nit-picky about spacing, and I'm going to slightly reword the messages so they match the others.

Thanks again!

PR #187, opened 2024-08-24
Fix bug in xUnit1051 to account for arbitrary argument order

### LINKED ISSUE #3011: xUnit1051 marks method with omitted optional parameters and named cancellationToken parameter
The last method call is still marked by the analyzer even though it is correctly passing in the cancellation token.

```c#
public class UnitTest1
{
    [Fact]
    public async Task Test()
    {
        await MethodWithCancellationToken(TestContext.Current.CancellationToken); // works
        await MethodWithCancellationToken(cancellationToken: TestContext.Current.CancellationToken); // works
        await MethodWithOptionalParametersCancellationToken(null, TestContext.Current.CancellationToken); // works
        await MethodWithOptionalParametersCancellationToken(a: null, cancellationToken: TestContext.Current.CancellationToken); // works

        await MethodWithOptionalParametersCancellationToken(cancellationToken: TestContext.Current.CancellationToken); // still marked by analyzer
    }

    private static Task MethodWithCancellationToken(CancellationToken cancellationToken = default) => Task.CompletedTask;
    private static Task MethodWithOptionalParametersCancellationToken(int? a = null, CancellationToken cancellationToken = default) => Task.CompletedTask;
}
```

campersau:
Argument order can not be trusted, always check if parameter matches argument parameter.

Fixes https://github.com/xunit/xunit/issues/3011

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/UseCancellationTokenTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/UseCancellationTokenTests.cs
@@ -53,9 +53,13 @@ public class UseCancellationTokenTests
 
 			class TestClass {{
 			    [Fact]
-			    public async Task TestMethod() {{
-			        await Task.Delay(1, {0});
+			    public void TestMethod() {{
+			        FunctionWithDefaults(42, {0});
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks!

PR #188, opened 2024-08-25
Add Fixer for xUnit1051 using TestContext.Current.CancellationToken

campersau:
Based on #187

After trying out v3 yesterday and fixing all cancellation tokens manually, I decided to write a fixer.


bradwilson:
Thanks!

PR #189, opened 2024-08-26
Feat trigger x1010 when InlineData is negative int and parameter is uint

NormanFrieman:
When pass negative int in inlinedata and the parameter of test is uint, the test will fail in runtime

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/InlineDataMustMatchTheoryParametersTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/InlineDataMustMatchTheoryParametersTests.cs
@@ -1038,17 +1038,18 @@ public class InlineDataMustMatchTheoryParametersTests
 				await Verify.VerifyAnalyzer(source);
 			}
 
-			[Fact]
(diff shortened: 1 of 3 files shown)


NormanFrieman:
@dotnet-policy-service agree

bradwilson:
Sorry for the delay, I thought I had already left a comment here but I must not have submitted it. ☹️ 

The fix provided here is very focused on specifically negative `int`s and `uint`s. An ideal fix would be generalized to support all integral types, ensuring compatibility.

NormanFrieman:
@bradwilson I've improved the test I created that tests negative int to uint conversion scenarios. Are you suggesting that I include a fix for this issue as well? [issue 2983](https://github.com/xunit/xunit/issues/2983)

bradwilson:
Thanks! I needed to fix some merge issues and wanted to add an edge case to the tests, but your branch is not open to pushes, so I had to do them myself. You're credited with the squashed commit: https://github.com/xunit/xunit.analyzers/commit/d25b3a43404a3ea0c11300b6cd0025accf2eebd8

PR #190, opened 2024-08-31
Fix xUnit1051 move argument list creation inside of change document

campersau:
Looks like in VS the change document part is run twice: once for the popup/preview and once for the actual execution. Which meant that the cancellation token got added twice when there wasn't a default one before.


bradwilson:
Perhaps I'm not understanding the bug, but I don't seem to be able to repro what I think you're suggesting is the bug:


https://github.com/user-attachments/assets/7a8fcf6b-d4a0-4ff9-8392-c769a3a664c0

campersau:
The bug happens when there are other optional parameters as well:

https://github.com/user-attachments/assets/fb5286b0-ed52-41e1-8f58-a4b0ee5edcb1

bradwilson:
~Thanks! I'll see if I can add a test for this as well.~

Never mind, I have a test for this but because it's only triggered by the preview (as you mentioned) I can't really test it in a unit test. 🙄 😂

PR #191, opened 2024-08-31
Support analyzers / fixers on custom Fact / Theory attributes

campersau:
There were two instances where the Fact / Theory attribute weren't checked with `IsAssignableFrom`.

The change in `CodeAnalysisExtensions.cs` allows analysers / fixers to work with custom attributes inheriting `FactAttribute` / `TheoryAttribute`.


bradwilson:
Sorry for the delay in getting to this.

We almost never report on (or fix) anything related to custom Fact and Theory attributes, because we can't anticipate what those attributes do or what their rules might be, and whether those rules will be in conflict with the rules we have.

Developers who are creating custom Fact and Theory attributes for third party consumption can choose to implement analyzer rules for their attributes (and they're welcome to copy our analyzer code/infrastructure when doing so).

As such, this is not a fix that we would take. Sorry about that.

PR #192, opened 2024-11-29
Handle switching from params to non-params overload in xUnit1051

### LINKED ISSUE #3068: xUnit1051 code fix chooses incorrect overload for CancellationToken
I ran into this while upgrading to v3 and with EF Core's FindAsync method inside my tests, which exposes a `params object[] keyValues` overload and an `object[] keyValues, CancellationToken cancellationToken` overload.

Given the following code:

```cs
public class UnitTest1
{
    [Fact]
    public async Task Test1()
    {
        await M(1);
    }
    
    private async Task M(params int[] numbers)
    {
        
    }

    private async Task M(int[] numbers, CancellationToken cancellationToken = default)
    {
        
    }
}
```

I get a diagnostic for xUnit1051 on the invocation of `M` in UnitTest1. But if I apply the code fix, I get 

```cs
[Fact]
public async Task Test1()
{
    await M(1, TestContext.Current.CancellationToken);
}
```

which is a compile error. The correct code is:

```cs
[Fact]
public async Task Test1()
{
    await M([1], TestContext.Current.CancellationToken);
}
```

There's an argument here that the EF Core API is not the best (ReSharper does warn me about the second method being hidden). In fact, it's a bit worse in that case since `params object[]` accepts the CancellationToken but just leads to a runtime failure by trying to treat it as a primary key.

That being said, I didn't see any existing issues with this so I thought I'd raise it all the same - maybe it's an easy fix.

SapiensAnatis:
If the analyzer has fired on a `params` method, it means that the new overload is actually array + `CancellationToken`, since C# forbids placing any argument after a params argument. To avoid generating invalid code, we need to replace the params expression with an array creation expression.

Does not currently handle the expanded range of allowed `params` types implemented in C# 13. Considering you can use _any_ type that implements `IEnumerable<T>`, I think it would be quite difficult to handle all the different cases there. Being a new feature, it's also less likely to be used, though this will of course change over time...

Closes [#3068](https://github.com/xunit/xunit/issues/3068)

### DIFF
--- a/src/xunit.analyzers.fixes/X1000/UseCancellationTokenFixer.cs
+++ b/src/xunit.analyzers.fixes/X1000/UseCancellationTokenFixer.cs
@@ -7,6 +7,7 @@ using Microsoft.CodeAnalysis.CodeActions;
 using Microsoft.CodeAnalysis.CodeFixes;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Operations;
 using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
 
 namespace Xunit.Analyzers.Fixes;
@@ -51,6 +52,9 @@ public class UseCancellationTokenFixer : BatchedCodeFixProvider
 		if (root.FindNode(diagnostic.Location.SourceSpan) is not InvocationExpressionSyntax invocation)
 			return;
 
+		if (semanticModel.GetOperation(invocation, context.CancellationToken) is not IInvocationOperation invocationOperation)
+			return;
+
 		var arguments = invocation.ArgumentList.Arguments;
 
 		for (var argumentIndex = 0; argumentIndex < arguments.Count; argumentIndex++)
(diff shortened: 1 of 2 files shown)


SapiensAnatis:
@dotnet-policy-service agree

bradwilson:
It looks like there are some trivial test failures to be fixed.

Normally I would fix these, but your branch is not open for me to contribute to. 😄

SapiensAnatis:
Oops! That should hopefully be fixed now. Thanks for letting me know.

It's odd you can't commit to my branch, that setting appears to be enabled :confused: 

image


edit: on the whitespace issue, I've noticed it's because the raw string literals use spaces for indentation but the rest of the code uses tabs. I've used tabs in my raw string literals, would you prefer these be switched to spaces?

bradwilson:
> edit: on the whitespace issue, I've noticed it's because the raw string literals use spaces for indentation but the rest of the code uses tabs. I've used tabs in my raw string literals, would you prefer these be switched to spaces?

I've tended to use spaces, because the way the fixers work by default in testing is the they appear to use the default settings (i.e., 4 spaces), so matching that default makes the results much more predictable.

Unfortunately, it's not easy to switch between tabs and spaces when editing the multi-line literal strings, which means I'm wearing out my spacebar. 😂

SapiensAnatis:
Makes sense. For consistency's sake I've adapted the new tests to do the same. Sorry you keep having to approve the workflows!

I assume the 4 space issue appears when a code fix adds a new line? Maybe you could fix it by passing an `.editorconfig` file that specifies to use tabs into the test? Looks like you could do this in `VerifyCodeFixV3` by setting `test.TestState.AnalyzerConfigFiles`: https://github.com/dotnet/roslyn-sdk/blob/35d5e46fd5c403194692c645d912a17d36ed74f5/tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.UnitTests/AnalyzerConfigFilesTests.cs#L38-L41

bradwilson:
> Maybe you could fix it by passing an `.editorconfig` file

I don't think I knew you could do this. 😂 I might do this and then fix everything to use tabs! Thanks!

bradwilson:
Thanks!

PR #193, opened 2025-04-23
Update xUnit1004 to not trigger with SkipWhen or SkipUnless

### LINKED ISSUE #2988: Update xUnit1004 to not trigger when using SkipWhen or SkipUnless
The purpose of xUnit1004 is to highlight that skipped tests should be temporary, but that's not true when using `SkipWhen` or `SkipUnless`, since that's a test that's conditionally skipped.

Update the analyzer to stop triggering when using one or the other.

joaonunomota:
Closes https://github.com/xunit/xunit/issues/2988

- Updates xUnit1004 analyzer to not trigger on `Fact` attributes with `SkipWhen` or `SkipUnless` parameters.
- The analyzer returns without reporting xUnit1004 if `attribute.ArgumentList is null`, as there cannot be a `Skip` parameter in the `Fact` attribute (consistent with previous behavior).
- The analyzer checks for `Skip`, `SkipWhen`, and `SkipUnless` in a single loop (for a slightly better performance) and returns without reporting xUnit1004 if `SkipWhen` or `SkipUnless` is found.
- The update to the analyzer does not report xUnit1004 in the case where `SkipWhen` or `SkipUnless` are explicitly mentioned, but have a `null` value. The explicit use of either parameter with a `null` value seems like a misuse of the parameters that would be better addressed by a separate analyzer / fixer.
- The new tests cover the cases for `Fact` attributes with `SkipWhen` or `SkipUnless` parameters and the existing `SkippedTest_Triggers` covers the cases without either.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/TestMethodShouldNotBeSkippedTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/TestMethodShouldNotBeSkippedTests.cs
@@ -1,4 +1,5 @@
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CSharp;
 using Xunit;
 using Verify = CSharpVerifier<Xunit.Analyzers.TestMethodShouldNotBeSkipped>;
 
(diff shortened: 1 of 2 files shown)


joaonunomota:
@bradwilson Could you give this a look when you get the chance, let me know if there is anything needed from me before that!

bradwilson:
Sorry, this fell off my list. I'll take a look right now!

bradwilson:
Thanks!

PR #195, opened 2025-07-06
Analyzer - theory data row inside theory data

snailtempo:
## Addresses [issue #3113](https://github.com/xunit/xunit/issues/3113) - Add analyzer to flag using ITheoryDataRow inside of TheoryData

Includes:
- Analyzer to detect `TheoryData<ITheoryDataRow>` and stronger types
- Added code fix to suggest` IEnumerable<ITheoryDataRow>` 
- Analyzer/Codefix tests

Note: This is my first time contributing to this project and also my first time writing a Roslyn analyzer, so I wasn’t completely sure about some of the design decisions. Feedback is very welcome!

For instance, I was not sure if a code fix should be provided in such a scenario:
```
TheoryData<ITheoryDataRow> GetData() { return null; } // Currently: Codefix enabled, but could lead to invalid code (depending on method implementation)
```


snailtempo:
@dotnet-policy-service agree

bradwilson:
Thanks! I'm consolidating the tests for better performance (the compilation process is extremely expensive, so I've been doing this to all the existing tests as time allows).

PR #196, opened 2025-09-13
xunit/xunit#1258 Analyzer for uninitialized member data

Vogel612:
Initial implementation of an analyzer that verifies `MemberData` has an associated initializer expression or static constructor.

This PR is in Draft to solicit feedback on test cases that still need covering and guidance on how to preferrably handle existing test cases generating new analyzer results now that there is a stricter analyzer.

### DIFF
--- a/src/xunit.analyzers/X1000/MemberDataShouldReferenceValidMember.cs
+++ b/src/xunit.analyzers/X1000/MemberDataShouldReferenceValidMember.cs
@@ -138,7 +138,7 @@ public class MemberDataShouldReferenceValidMember : XunitDiagnosticAnalyzer
 				if (!memberSymbol.IsStatic)
 					ReportNonStatic(context, attributeSyntax, memberProperties);
 
-				if (!IsInitialized(memberSymbol, compilation, context))
+				if (!IsInitialized(memberSymbol, context))
 					ReportMemberMustBeWrittenTo(context, memberSymbol);
 
 				// Unwrap Task or ValueTask, but only for v3
@@ -201,7 +201,6 @@ public class MemberDataShouldReferenceValidMember : XunitDiagnosticAnalyzer
 
 	static bool IsInitialized(
 		ISymbol memberSymbol,
-		Compilation compilation,
 		SyntaxNodeAnalysisContext context
 	)
 	{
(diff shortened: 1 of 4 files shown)


Vogel612:
@dotnet-policy-service agree

src/xunit.analyzers/X1000/MemberDataShouldReferenceValidMember.cs
```diff
@@ -195,6 +199,43 @@ memberReturnType is INamedTypeSymbol namedMemberReturnType &&
...
+			return true;
+
+		var declarationContainer = declarationReference.SyntaxTree.GetCompilationUnitRoot();
+		var staticConstructors = declarationContainer.DescendantNodes()
+			.OfType<MethodDeclarationSyntax>()
+			.Where(mds => semantics.GetDeclaredSymbol(mds)?.MethodKind == MethodKind.StaticConstructor);
+
+		foreach (var ctor in staticConstructors)
+		{
+			var analysis = semantics.AnalyzeDataFlow(ctor);
```
Vogel612: this setup is unfortunately insufficient to correctly determine whether a member is assigned in a static constructor.

Vogel612:
@bradwilson Would appreciate some feedback on this. As a note: this was developed on linux, and I haven't had the chance to do a verification run on windows

bradwilson:
Sorry for the delay getting to this, I haven't really had much personal time for xUnit.net in the last month.

bradwilson:
Thanks!

bradwilson:
Something odd is happening: this analyzer will report with the compiler, but does not report inside Visual Studio.

Using this as my test:

```csharp
public class UnitTest1
{
    public static TheoryData<int> DataSource;

    [Theory]
    [MemberData(nameof(DataSource))]
    public void TestMethod(int x)
    {
        Assert.NotEqual(0, x);
    }
}
```

I get nothing in Visual Studio:

image

But when I build, the warning shows:

`UnitTest1.cs(3,35,3,45): warning xUnit1053: The member DataSource referenced by MemberData is not initialized before use. Add an inline initializer or initialize the value in the static constructor. (https://xunit.net/xunit.analyzers/rules/xUnit1053)`

I've never seen anything like this, and none of the other analyzers are affected. Is there any reason you can think of that your analyzer implementation doesn't work inside Visual Studio?

Vogel612:
@bradwilson honestly no idea why that difference would show up like that. If I was forced to guess, I would try to figure out whether the same version of roslyn is used between the two analysis modes, but from the little experience I have with writing analyzers, these should behave the same...

PR #197, opened 2025-11-20
Fix X2022: Preserve all arguments when fixing negated boolean assertions

### LINKED ISSUE #3446: xUnit2022 code fixer removes the second argument
Repro:

```c#
namespace TestProject1;

public class UnitTest1
{
    [Fact]
    public void Test1()
    {
        Assert.True(!false, userMessage: "");
    }
}
```

The code fixer removes the second argument (`userMessage`).

Image

Tenno641:
Fixes xunit/xunit#3446 - Code fixer was incorrectly removing second argument (userMessage) when converting Assert.True(!false) to Assert.False(false)

- Modified BooleanAssertsShouldNotBeNegatedFixer to preserve all arguments
- Added tests for userMessage, multiple arguments, and named arguments
- Existing test case continues to work as before


Tenno641:
@dotnet-policy-service agree

bradwilson:
Thanks for the PR!

The new tests are not passing in the automated PR build (failing in all three OSes, so it doesn't seem to be OS-specific).

Tenno641:
Hey @bradwilson 
Sorry for the mess with this PR, I couldn’t fix the failing test yet, so I’ll close it for now and try again later. Thanks for your patience!

PR #198, opened 2025-11-21
Fix preserve arguments

### LINKED ISSUE #3446: xUnit2022 code fixer removes the second argument

Tenno641:
Fixes xunit/xunit#3446 
#### Code fixer was incorrectly removing second argument (userMessage) when converting Assert.True(!false, userMessage: "") to Assert.False(false)

- Modified BooleanAssertsShouldNotBeNegatedFixer to preserve all arguments 
- Added tests to ensure userMessage (both positional and named) is preserved by the code fix.

Note: This is a clean version of my earlier PR, which I closed due to some git issues. Sorry about the noise.

### DIFF
--- a/src/xunit.analyzers.fixes/X2000/BooleanAssertsShouldNotBeNegatedFixer.cs
+++ b/src/xunit.analyzers.fixes/X2000/BooleanAssertsShouldNotBeNegatedFixer.cs
@@ -66,9 +66,7 @@ public class BooleanAssertsShouldNotBeNegatedFixer : XunitCodeFixProvider
 
 				var newArguments = new List<ArgumentSyntax> { newFirstArgument };
 				if (originalArguments.Count > 1)
-				{
 					newArguments.AddRange(originalArguments.Skip(1));
-				}
 
 				editor.ReplaceNode(
 					invocation,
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks!

-----

FWIW, you will note that I rewrote all the tests so as to use a single acceptance test rather than using multiple tests and theories. I have determined that the cost of compilation for each test (or each data row of a data-driven test) is overwhelmingly slowing down the overall developer turn-around time, so I'm preferring singular acceptance tests with copy/pasted test scenarios in order to make everything run faster.

On my developer machine (AMD 9950X w/ 16 cores, 64GB of RAM) each individual compilation step required costs on the order of hundreds of milliseconds, so in this case assuming each test takes 200ms, the switch from 11 tests to 1 saves 2 seconds per test assembly, so 8 seconds of overall CPU time. The overall savings in the CI environment is even more valuable, since those machines are anemically underpowered.

This is an ongoing effort. Right now my machine runs all 4 test projects in parallel using 295 seconds of CPU time (and 81.5 seconds of "clock time"). I'm hoping that when I'm done I'll be well under half that time, so it helps you, helps me, and especially helps me stop wasting CI time on a shared resource. 😄

PR #199, opened 2025-12-23
Fix X1019: Added special-cases handling to correctly validate types convertible to IEnumerable<T[]>

### LINKED ISSUE #3411: xUnit1019 false positive
```cs
public sealed class ValidExamples : IEnumerable<string[]>
{
    private readonly List<string[]> _items = [];

    public void Add([StringSyntax("GraphQL")] string sdl)
    {
        _items.Add([sdl]);
    }

    public IEnumerator<string[]> GetEnumerator()
    {
        return _items.GetEnumerator();
    }

    IEnumerator IEnumerable.GetEnumerator()
    {
        return GetEnumerator();
    }
}
```

```cs
// Valid
IEnumerable<object[]> x = new ValidExamples();
```

Image

It runs without issues.

fnipo:
Fixes [xunit/xunit#3411](https://github.com/xunit/xunit/issues/3411)

### Added special-cases handling to MemberDataShouldReferenceValidMember.cs

- Extracts the generic argument `T` from `IEnumerable<T>` and treats any array type (`T[]`) as assignable to `object[]`. This enables `IEnumerable<string[]>` to be recognized as a valid MemberData reference type.
- Adds support for named types that are convertible to `IEnumerable<T[]>` by analyzing their source‑declared type syntax. As a result, classes declared such as `class NamedTypeForIEnumerableStringArray : IEnumerable<string[]>` are now correctly identified as valid MemberData reference types.

Alternatives explored:
- While evaluating `compilation.ClassifyConversion(fromType, toType)`, I found that Roslyn often reports `IsImplicit = true` for conversions that are not actually valid at runtime. Because of this, it isn’t a reliable mechanism for determining true assignability or convertibility between types.
- When resolving named types, I wasn’t able to reliably extract the concrete generic argument using only `ITypeSymbol` / `INamedTypeSymbol`. For example, in the case of `NamedTypeForIEnumerableStringArray` class, the implemented interface appeared as `IEnumerable<T>` rather than `IEnumerable<string[]>`, which prevents inspection of the actual generic type argument. To address this, I leveraged the source‑declared type syntax instead, allowing me to accurately extract the generic argument and walk the declared base‑type chain.

### Notes
I’d welcome any feedback on this, as it’s my first contribution to this repository. Some of the logic I introduced could potentially be folded into `SymbolExtensions.IsAssignableFrom()` to improve its handling of special‑case assignability scenarios, particularly where non‑exact matches are interesting. However, since that method is widely used across the codebase, I opted to implement the special‑case handling locally to avoid unintended regressions, especially given my limited familiarity with the repo.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/MemberDataShouldReferenceValidMemberTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/MemberDataShouldReferenceValidMemberTests.cs
@@ -1,5 +1,3 @@
-using System.Collections;
-using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.CSharp;
 using Xunit;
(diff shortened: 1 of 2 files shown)


fnipo:
@dotnet-policy-service agree

bradwilson:
Thanks!

PR #200, opened 2026-01-23
Fix AD0001 in MemberDataShouldReferenceValidMember for metadata symbols

### LINKED ISSUE #3474: Analyzer crashes for inherited `[MemberData]` from referenced assembly
## Summary

The `Xunit.Analyzers.MemberDataShouldReferenceValidMember` analyzer throws an exception when `[MemberData]` references a static member inherited from a base class defined in a different assembly. Tests still run and pass, but the build produces an AD0001 analyzer warning.

## Environment

* OS: macOS (arm64)
* .NET: 8.0
* xunit.v3.assert: 3.2.2
* xunit.analyzers: 1.27.0

## Reproduction steps

Example Project structure:

* `Data` project: Contains a base test class with a static member data source (`BaseMemberData.cs`).

* `Tests` project: References `Data`. Contains a test class that inherits from the base class and uses
  `[MemberData(nameof(SharedData))]` from the other assembly.

Run `dotnet test` in `Tests`

## Observed behavior

During the build, the analyzer crashes, triggering the following warning:

```
CSC : warning AD0001: Analyzer 'Xunit.Analyzers.MemberDataShouldReferenceValidMember' threw an exception of type 'System.InvalidOperationException' with message 'Sequence contains no elements'.
```

The tests themselves execute and pass. The failure occurs only during analyzer execution.

## Expected behavior

The analyzer should not throw an exception.

Tenno641:
Fixes xunit/xunit#3474

Prevent a crash in `MemberDataShouldReferenceValidMember` when analyzing metadata symbols from referenced projects by guarding against empty `DeclaringSyntaxReferences`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -409,3 +409,4 @@ FodyWeavers.xsd
 
 # JetBrains Rider
 *.sln.iml
+.idea/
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks!

PR #202, opened 2026-02-08
Fix xUnit1042 false positive: Increase TheoryData arity limit to 20

### LINKED ISSUE #3485: xUnit1042 false positive for TheoryData with more than 10 arguments
```csharp
public static class Test
{
    // No issues
    [Theory]
    [MemberData(nameof(TestData.TheoryData10Params), MemberType = typeof(TestData))]
    public static void Test10Params(int a, int b, int c, int d, int e, int f, int g, int h, int j, int k)
    {
        Assert.Equal(45, a + b + c + d + e + f + g + h + j + k);
    }

    // False-positive xUnit1042: The member referenced by the MemberData attribute returns untyped data rows, such as object[].
    // Consider using TheoryData<> or IEnumerable<TheoryDataRow<>> as the return type to provide better type safety.
    [Theory]
    [MemberData(nameof(TestData.TheoryData11Params), MemberType = typeof(TestData))]
    public static void Test11Params(int a, int b, int c, int d, int e, int f, int g, int h, int j, int k, int l)
    {
        Assert.Equal(55, a + b + c + d + e + f + g + h + j + k + l);
    }
}

public static class TestData
{
    public static TheoryData<int, int, int, int, int, int, int, int, int, int> TheoryData10Params =>
        new() { { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 } };
    public static TheoryData<int, int, int, int, int, int, int, int, int, int, int> TheoryData11Params =>
        new() { { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 } };
}
```
xunit.v3 3.2.2
net10.0 (Win11)

hajdu-patrik:
## Description
This PR resolves an issue where the analyzer would incorrectly report `xUnit1042` and `xUnit1037` for `TheoryData` implementations utilizing more than 10 generic arguments.

The `TypeSymbolFactory` was previously hardcoded to verify `TheoryData` and `TheoryDataRow` generic arity only up to 10. I have increased this limit to 20 to support larger datasets, consistent with modern C# capabilities.

## Related Issue
Fixes xunit/xunit#3485

## Changes
- Updated `TypeSymbolFactory.TheoryData_ByGenericArgumentCount` loop limit to 20.
- Updated `TypeSymbolFactory.TheoryDataRow_ByGenericArgumentCount_V3` loop limit to 20.
- Added a regression test in `MemberDataShouldReferenceValidMemberTests` to verify 11 parameters are now accepted.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/MemberDataShouldReferenceValidMemberTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/MemberDataShouldReferenceValidMemberTests.cs
@@ -1452,6 +1452,30 @@ public class MemberDataShouldReferenceValidMemberTests
 		}
 
 		[Fact]
+		public async ValueTask TheoryData_With11Parameters_ShouldNotTrigger1042()
+		{
+			var source = /* lang=c#-test */ """
+				#pragma warning disable xUnit1053
+				using Xunit;
+
+				// IMPORTANT: namespace Xunit block should be used so that the Factory can find it!
+				namespace Xunit 
+				{
+					public class TheoryData<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11> : TheoryData { }
+				}
+
+				public class TestClass {
+					public static TheoryData<int, int, int, int, int, int, int, int, int, int, int> Data11;
+
+					[MemberData(nameof(Data11))]
+					public void TestMethod(int a, int b, int c, int d, int e, int f, int g, int h, int i, int j, int k) { }
+				}
+				""";
+
+			await Verify.VerifyAnalyzer(source);
+		}
+
+		[Fact]
 		public async Task V3_only()
 		{
 			var source = /* lang=c#-test */ """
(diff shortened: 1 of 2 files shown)


hajdu-patrik:
@dotnet-policy-service agree

bradwilson:
This will need to be updated to be version-sensitive.

* In xUnit.net v2, `TheoryData<>` supported up to 10 items
* In xUnit.net v3, `TheoryData<>` supports up to 15 items

The change occurred here: https://github.com/xunit/xunit/commit/923310ab07598fec7a1ce4fd6872693485626ebf

It was shipped in a prerelease version of v3, which is not something that needs to be accounted for. It is reasonable in the analyzer to assume that any version of v3 supports up to 15 values and any version of v2 supports up to 10 values.

bradwilson:
(Technically, we moved from 5 to 10 in v2 2.3, but that's old enough that it shouldn't matter at this point.)

hajdu-patrik:
Oh than sorry for the bad contribution. And thanks for the information!

PR #204, opened 2026-02-18
Improves handling of generic type constraints and CRTP (xunit/xunit#2986)

### LINKED ISSUE #2986: Analyzer for xUnit1010 is weird when generic method is constraint with CRTP-types
The xUnit1010 analyzer exhibits some unexpected behaviour when using generic methods, where the constraint makes use of the [curiously recurring template pattern](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern).

It creates an error for every `[InlineData]` attribute and ignores even faulty `[MemberData]` attributes

```csharp
public class GenericTypeConstraints
{
    public static TheoryData<uint> Unsigned => new([0, 1, 2, 3]);
    public static TheoryData<int> Signed => new([0, 1, 2, 3]);

    [Theory]
    [InlineData(-1)]
    [InlineData(0U)]
    [MemberData(nameof(Unsigned))]
    [MemberData(nameof(Signed))]
    public void UnsignedNumberIsAtLeastZero<T>(T number)
        where T : IUnsignedNumber<T> => Assert.False(T.IsNegative(number));
}
```

**Expected bahaviour:**
The `[MemberData(nameof(Signed))]`  and `[InlineData(-1)]` attributes should create an error as `int` does not implement `IUnsignedNumber<int>`.
The `[InlineData(0u)]` and  `[MemberData(nameof(Unsigned))]` should be fine.

**Actual behaviour:**
For `[InlineData(-1)]` and `[InlineData(0ul)]` the analyzer emits xUnit1010 errors:

> The value is not convertible to the method parameter 'number' of type 'T'. Use a compatible data value. (https://xunit.net/xunit.analyzers/rules/xUnit1010)

It does not emit any messages for either of the `MemberData` attributes.

ssaporito:
Partially fixes [this issue](https://github.com/xunit/xunit/issues/2986) by improving handling of generic type constraints, including CRTP. This only applies to InlineData, since MemberData is not presently checked for constraints. Includes a new unit test built around the initial issue.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/InlineDataMustMatchTheoryParametersTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/InlineDataMustMatchTheoryParametersTests.cs
@@ -988,6 +988,32 @@ public class InlineDataMustMatchTheoryParametersTests
 				await Verify.VerifyAnalyzer(source);
 			}
 
+#if ROSLYN_LATEST && NET8_0_OR_GREATER
+			[Fact]
+			public async Task TypeConstraint_WithCRTP() // CRTP: Curiously Recurring Template Pattern or T: Interface<T>
+			{
+				var source = /* lang=c#-test */ """
+				using Xunit;
+				using System.Numerics;
+
+				public class TestClass {
+					[Theory]
+					[InlineData(0U)]
+					[InlineData(2U)]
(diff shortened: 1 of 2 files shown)


ssaporito:
@dotnet-policy-service agree

bradwilson:
Thanks!

PR #205, opened 2026-02-27
Fix xUnit1015 when MemberData is declared on derived type

### LINKED ISSUE #3501: Update xUnit1015 to look on derived classes when base class is abstract
Today, xUnit1015 will trigger for [MemberData] when the member only exists on the derived type:

```csharp
public abstract class BaseClassWithTestWithoutData
{
    [Theory]
    // vvv == xUnit1015 triggers here == vvv
    [MemberData(nameof(SubClassWithTestData.TestData))]
    public void Test(int x)
    { }
}

public class SubClassWithTestData : BaseClassWithTestWithoutData
{
    public static IEnumerable<object?[]> TestData()
    {
        yield return new object?[] { 42 };
    }
}
```

This could be updated not to trigger after validating that all derived classes implement the required member. This will probably at least warrant updating the warning message, and may end up just omitting xUnit1015 and instead writing a new rule that specifically looks for this. The implementer can decide which is best.

AwsAqh:
Fixes xunit/xunit#3501

- Suppress xUnit1015 when the test is declared on an abstract base class and the referenced MemberData member exists on all derived concrete classes in the compilation.
- Added regression test covering the abstract-base + derived MemberData scenario.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/MemberDataShouldReferenceValidMemberTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/MemberDataShouldReferenceValidMemberTests.cs
@@ -2,7 +2,7 @@ using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.CSharp;
 using Xunit;
 using Verify = CSharpVerifier<Xunit.Analyzers.MemberDataShouldReferenceValidMember>;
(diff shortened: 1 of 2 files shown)


AwsAqh:
@dotnet-policy-service agree
 @dotnet-policy-service agree company="Microsoft"

bradwilson:
Thanks!

PR #209, opened 2026-04-23
Add xUnit1054: Timeout tests should reference TestContext.Current.CancellationToken

### LINKED ISSUE #2977: Create v3 analyzer for `Timeout` tests which don't use `TestContext.Current.CancellationToken`
See: https://github.com/xunit/xunit/issues/2222#issuecomment-2176452777

Note that although we did just [add an analyzer](https://github.com/xunit/xunit.analyzers/commit/2188a2df0d7ad684dd10fefdd250973b998b5aee) to flag usages of APIs from v3 tests where a `CancellationToken`, this analyzer is slightly different. The intended purpose here would be search any test with a `Timeout`, and make sure there is at least **_one_** reference, anywhere, to `TestContext.Current.CancellationToken`. This is not about inspecting the individual APIs that are called, but rather to make sure that somewhere in the body of the test, something is aware that tests with timeouts must be paying attention to the cancellation token so that they can terminate as quickly as possible when they're running too long.

KitKeen:
Fixes xunit/xunit#2977.

Adds a new v3-only analyzer `xUnit1054` that flags any test method annotated with a `Fact`-derived attribute and a non-zero `Timeout` whose body contains no reference to `TestContext.Current.CancellationToken`. Per the issue, tests with a timeout should be paying attention to the cancellation token so they can terminate promptly when the timeout is exceeded — this surfaces the omission at build time.

## What the analyzer does
- Runs only when v3 references are present (`XunitV3DiagnosticAnalyzer`).
- For each method symbol with an attribute whose attribute class derives from `FactAttribute` and whose `Timeout` named argument is a positive integer, walks the method's operation blocks (including lambdas and local functions) looking for any `IPropertyReferenceOperation` that resolves to `Xunit.TestContext.CancellationToken`.
- If none is found, reports at the `Timeout = ...` attribute argument (falling back to the attribute / method location).
- Per the issue, "reference anywhere in the body" is sufficient — we don't inspect which APIs receive the token, only that the test body is aware of it. This matches the complementary `xUnit1051` which handles the per-invocation case.

## Why at the attribute-argument location
The violation is about the test's contract (it promised to respect `Timeout`), so the squiggle should point at `Timeout = 1000` rather than the method body. Method location is used as a fallback for custom attribute application shapes.

## What's out of scope
- Custom `Fact`-derived attributes that set `Timeout` in the constructor rather than as a named argument at the call site. The analyzer only inspects the syntactic `NamedArguments` on the attribute application, mirroring how `Timeout` is normally authored. Happy to extend if the maintainers prefer.
- Class-level capture of `TestContext.Current.CancellationToken` into a field. The issue describes the check as "somewhere in the body of the test", so per-method scope seemed correct; again, happy to broaden.

## Tests
`TestMethodWithTimeoutShouldUseCancellationTokenTests` covers:
- No `Timeout` → no diagnostic.
- `Timeout = 0` → no diagnostic.
- `Timeout = 1000` with direct / via-local / in-lambda references to `TestContext.Current.CancellationToken` → no diagnostic.
- `[Fact(Timeout = 1000)]` with no reference → diagnostic.
- `[Theory(Timeout = 1000)]` with no reference → diagnostic.
- `[MyFact(Timeout = 1000)]` (derived attribute) with no reference → diagnostic.
- Non-test method without any reference → no diagnostic.

## Checklist
- [x] I've read the guidelines for contributing.
- [ ] I've posted a comment on an issue with a detailed description of how I am planning to contribute and got approval from a member of the team. *(Opening this PR as the planned approach — happy to discuss on the issue first if preferred; the issue itself has a clear spec and no open questions, so it seemed lighter-weight to propose in code.)*
- [ ] The code builds and tests pass locally. *(Could not verify locally — my machine only has .NET SDK 8 installed and the repo requires 10. Relying on CI.)*
- [x] Commit message follows the repo format.
- [x] Tests for the change have been added.
- [x] Code follows the same patterns and style as existing code in this repo.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/TestMethodWithTimeoutShouldUseCancellationTokenTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/TestMethodWithTimeoutShouldUseCancellationTokenTests.cs
@@ -1,0 +1,180 @@
+using System.Threading.Tasks;
+using Xunit;
+using Verify = CSharpVerifier<Xunit.Analyzers.TestMethodWithTimeoutShouldUseCancellationToken>;
+
+public class TestMethodWithTimeoutShouldUseCancellationTokenTests
+{
+	[Fact]
+	public async Task NoTimeout_DoesNotTrigger()
+	{
+		var source = /* lang=c#-test */ """
+			using System.Threading;
+			using Xunit;
+
+			public class TestClass {
+				[Fact]
+				public void TestMethod() {
+					Thread.Sleep(1);
+				}
+			}
+			""";
+
+		await Verify.VerifyAnalyzerV3(source);
+	}
+
+	[Fact]
+	public async Task ZeroTimeout_DoesNotTrigger()
+	{
+		var source = /* lang=c#-test */ """
+			using System.Threading;
+			using Xunit;
+
+			public class TestClass {
+				[Fact(Timeout = 0)]
+				public void TestMethod() {
+					Thread.Sleep(1);
+				}
+			}
+			""";
+
+		await Verify.VerifyAnalyzerV3(source);
+	}
+
+	[Fact]
+	public async Task TimeoutWithDirectCancellationTokenReference_DoesNotTrigger()
+	{
+		var source = /* lang=c#-test */ """
+			using System.Threading.Tasks;
+			using Xunit;
+
+			public class TestClass {
+				[Fact(Timeout = 1000)]
+				public async Task TestMethod() {
+					await Task.Delay(1, TestContext.Current.CancellationToken);
+				}
+			}
+			""";
+
+		await Verify.VerifyAnalyzerV3(source);
+	}
+
+	[Fact]
+	public async Task TimeoutWithCancellationTokenViaLocal_DoesNotTrigger()
+	{
+		var source = /* lang=c#-test */ """
+			using System.Threading.Tasks;
+			using Xunit;
+
+			public class TestClass {
+				[Fact(Timeout = 1000)]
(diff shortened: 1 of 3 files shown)


bradwilson:
Thanks!

Unfortunately the `main` branch is extremely out of date as we push towards xunit.analyzers 2.0 (to be co-released with core framework 4.0). xUnit1054 is already in use on the `rel/2.0.0` branch: https://xunit.net/xunit.analyzers/rules/xUnit1054

There's more complications here as well with Native AOT. For example, `FactAttribute` is sealed in Native AOT, so you can't just look for things that are "derived from `FactAttribute`" in analyzers; instead, the `XunitContext.Core.FactAndTheoryAttributeTypes` collection contains all the appropriate `FactAttribute` attribute types (Fact, CulturedFact, Theory, CulturedTheory) that would need to be checked against. A quick search for `FactAndTheoryAttributeTypes` in the `rel/2.0.0` branch can show instances of how this is generally used.

I'm happy to do the work of moving it to `rel/2.0.0`, though obviously the existing PR would not be appropriate, since it currently targets `main` (and you're definitely going to have conflicts when you try to do that, given that xUnit1054 already exists).

Would you like to attempt to rework this for `rel/2.0.0` or would you like me to do it?

bradwilson:
I'm going to do the retargeting so that I can get this merged in advance of the next prerelease this weekend.

bradwilson:
Retargeted and merged here: https://github.com/xunit/xunit.analyzers/commit/51b44984a1de8ee2dd2c7eea43b213e24f749a77

Thanks!

PR #210, opened 2026-04-29
Improve xUnit1045 to suppress diagnostic when all TheoryData values are provably serializable

ana1250:
Fixes #2981

### Problem

`xUnit1045` fires for `TheoryData<object>` (and other `PossiblySerializable` type arguments) even when the collection is initialized exclusively with serializable values like primitives, strings, or constants:

```csharp
public static TheoryData<object> Data => new()
{
    { 1 },
    { "hello" },
};

[Theory]
[MemberData(nameof(Data))] // xUnit1045 fires here — false positive
public void Test1(object value) { }
```

### Solution

When the analyzer encounters a `PossiblySerializable` type argument (like `object`), it now performs **value-level inspection** on the data source before reporting `xUnit1045`. It navigates from the `[MemberData]` attribute to the backing member's declaration via `ISymbol.DeclaringSyntaxReferences`, extracts the collection initializer, and checks the compile-time type of every value using the semantic model.

If **all** values are `AlwaysSerializable` (e.g., `int`, `string`, `bool`, `DateTime`, etc.), the diagnostic is suppressed. Otherwise, it fires as before.

> **Note:** This only applies to `xUnit1045` (`PossiblySerializable`). `xUnit1044` (`NeverSerializable`) is never suppressed by value-level inspection, since the type argument itself is definitively non-serializable regardless of values.

### What's supported

- **Fields**: `public static readonly TheoryData<object> Data = new() { { (object)1 } };`
- **Properties**: expression-bodied, initializer-based, and single-return getter bodies
- **Methods**: expression-bodied and single-return methods
- **Cast unwrapping**: `(object)1` is unwrapped to inspect the underlying `int`
- **Constants**: `const int X = 1; new TheoryData<object> { (object)X }` — semantic model resolves the type
- **Null literals**: treated as serializable (xUnit handles null)

### What correctly falls back to warning

- **Empty initializers** `{ }` — can't prove safety, collection may be mutated elsewhere
- **`ClassData`** — can't inspect class internals
- **Data source in a different file** — avoids `RS1030` (`Compilation.GetSemanticModel()` is forbidden in analyzers); also covers partial classes where the member is in a separate file
- **Complex method bodies** — anything beyond a single return/expression body
- **Non-serializable values** — `new SomeClass()` correctly keeps the diagnostic

### Known limitation

The analysis is based on the collection initializer at declaration time. If a `TheoryData<object>` field is initialized with serializable values but later mutated via `Add()` with non-serializable data, the diagnostic will not fire. This is an inherent trade-off of compile-time analysis — the same limitation exists for other initializer-based inspections, and is analogous to the existing `DisableDiscoveryEnumeration` opt-out which also trusts the developer's intent.

### Changes

- `TheoryDataTypeArgumentsShouldBeSerializable.cs` — Added `CanStaticallyVerifyAllValuesAreSerializable`, `IsExpressionAllSerializable`, and `IsValueSerializable` methods. Refactored `TheoryDataTypeArgumentFinder` to expose the data source `ISymbol` alongside the type arguments.
- `X1045_TheoryDataTypeArgumentsShouldBeSerializableTests.cs` — Added `SafeObjectClass` (suppression) and `UnsafeObjectClass` (still fires) test scenarios inline with existing test structure.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/X1045_TheoryDataTypeArgumentsShouldBeSerializableTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/X1045_TheoryDataTypeArgumentsShouldBeSerializableTests.cs
@@ -22,9 +22,9 @@ public class X1045_TheoryDataTypeArgumentsShouldBeSerializableTests
 
 			public class ObjectClass {
 				public sealed class Class : TheoryData<object> { }
-				public static readonly TheoryData<object> Field = new TheoryData<object>() { };
-				public static TheoryData<object> Method(int a, string b) => new TheoryData<object>() { };
-				public static TheoryData<object> Property => new TheoryData<object>() { };
+				public static readonly TheoryData<object> Field = new TheoryData<object>() { new object() };
+				public static TheoryData<object> Method(int a, string b) => new TheoryData<object>() { new object() };
+				public static TheoryData<object> Property => new TheoryData<object>() { new object() };
 
 				[Theory]
 				[MemberData(nameof(Field), DisableDiscoveryEnumeration = true)]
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks!

PR #211, opened 2026-05-22
Fix InlineData parameter index after invalid literal

### LINKED ISSUE #3569: xUnit1010: Invalid literal causes invalid reporting
Hello!
I get xUnit1010 error when parameters in test decimal and Enum.

Simple code:

```csharp
public enum MyEnum
{
    Value1,
    Value2,
    Value3
}

public class UnitTest1
{
    [Theory]
    [InlineData(1.1m, MyEnum.Value1)]
    public void Test1(decimal param1, MyEnum param2)
    {
        Assert.True(true);
    }
}
```

Error displays on enum parameter but text referenced to first:
> The value is not convertible to the method parameter 'param1' of type 'decimal'. Use a compatible data value.

Image

wsgaero:
## Summary
- Advances the theory parameter index after an invalid InlineData literal so subsequent values stay aligned with their parameters
- Adds a regression test for the decimal literal + enum parameter scenario from xunit/xunit#3569

Fixes xunit/xunit#3569

## Tests
- `dotnet test src/xunit.analyzers.tests/xunit.analyzers.tests.csproj --filter FullyQualifiedName~InvalidLiteralDoesNotDesynchronizeSubsequentParameters --no-restore`
- `dotnet test src/xunit.analyzers.tests/xunit.analyzers.tests.csproj --filter FullyQualifiedName~X1010_IncompatibleValueType --no-restore`

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/InlineDataMustMatchTheoryParametersTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/InlineDataMustMatchTheoryParametersTests.cs
@@ -332,6 +332,30 @@ public class InlineDataMustMatchTheoryParametersTests
 
 			await Verify.VerifyAnalyzer(source);
 		}
+
+		// https://github.com/xunit/xunit/issues/3569
+		[Fact]
+		public async Task InvalidLiteralDoesNotDesynchronizeSubsequentParameters()
+		{
+			var source = /* lang=c#-test */ """
+				using Xunit;
+
+				public enum MyEnum {
+					Value1,
+					Value2,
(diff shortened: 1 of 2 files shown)


wsgaero:
@dotnet-policy-service agree

bradwilson:
Thanks!

I modified the change slightly and ported it to the `rel/2.0.0` branch, and committed it here: https://github.com/xunit/xunit.analyzers/commit/80d1379ad08f31d1e44a570574a14e72158b27fb?w=1

The fix is available in `2.0.0-pre.54` https://xunit.net/docs/using-ci-builds

PR #212, opened 2026-06-18
xunit/xunit#3336: Fix xUnit2023 fixer trivia and indentation

### LINKED ISSUE #3336: Issues with the fixer for xUnit2023
**Issue 1**

```cs
        // assert
        Assert.Collection(list, s => Assert.Equal(_strings[0], s));
```

Is fixed to:

```cs
        // assert
        var s = // assert
Assert.Single(list);
        Assert.Equal(_strings[0], s);
```

(duplicated comment, missing indentation)

**Issue 2**

Many cases are flagged but the fixer doesn't do anything. For example:

```cs
        Assert.Collection(result.Errors,
            error => Assert.Equal(
                "Response status code does not indicate success: 404 (Not Found).",
                error.Message));

        Assert.Collection(
            type.Locations.AsEnumerable(),
            t => Assert.Equal(DirectiveLocation.Query, t));

        Assert.Collection(
            query.Fields["address"].Directives,
            item => Assert.Equal(FederationTypeNames.ExternalDirective_Name, item.Type.Name));
```

(there are much more complicated cases, though)

**Issue 3**

```cs
        Assert.True(success1);
        Assert.Collection(registered1, t => Assert.Equal("abc", t.Id));
        Assert.False(success2);
        Assert.Collection(registered2, t => Assert.Equal("abc", t.Id));
```

Is fixed to:

```cs
        Assert.True(success1);
        var t = Assert.Single(registered1);
        Assert.Equal("abc", t.Id);
        Assert.False(success2);
        var t = Assert.Single(registered2);
        Assert.Equal("abc", t.Id);
```

(variable `t` is used twice ... not as big a deal, but could possibly have a numeric suffix added/incremented)

LindqvistMartin:
Fixes xunit/xunit#3336.

The fixer reused the original invocation and lambda nodes without clearing their trivia, so a `// comment` on the `Assert.Collection` got copied onto the generated `var` line (with its indentation reset to column zero), and a multi-line inspector body kept its old indentation instead of lining up under the new statement.

Stripping the trivia off the reused invocation fixes the comment; normalizing the inspector body fixes the indentation. Added a regression test for a leading comment, a multi-line collection, and a multi-line inspector.

### DIFF
--- a/src/xunit.analyzers.fixes/X2000/AssertSingleShouldBeUsedForSingleParameterFixer.cs
+++ b/src/xunit.analyzers.fixes/X2000/AssertSingleShouldBeUsedForSingleParameterFixer.cs
@@ -43,7 +43,7 @@ public class AssertSingleShouldBeUsedForSingleParameterFixer : XunitCodeFixProvi
 	static IEnumerable<SyntaxNode> GetLambdaStatements(SimpleLambdaExpressionSyntax lambdaExpression)
 	{
 		if (lambdaExpression.ExpressionBody is InvocationExpressionSyntax lambdaBody)
-			yield return ExpressionStatement(lambdaBody).WithAdditionalAnnotations(Formatter.Annotation, Simplifier.Annotation);
+			yield return ExpressionStatement(lambdaBody.NormalizeWhitespace()).WithAdditionalAnnotations(Formatter.Annotation, Simplifier.Annotation);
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks! I ported your changes to the `rel/2.0.0` branch (since that's where the next release is coming from, and the source has changed dramatically from `main`): https://github.com/xunit/xunit.analyzers/commit/d81d5b9144ad1ab47ba0acf57ce41dfbe2fb2257

PR #213, opened 2026-07-05
Report xUnit1025 for InlineData duplicated via implicit numeric conversion

KitKeen:
Fixes xunit/xunit#2957

xUnit1025 compared InlineData rows using the raw typed constants, so two rows that produce the same value for a parameter through an implicit numeric conversion (for example `1` and `(byte)1` for an `int` parameter) were not reported as duplicates.

This change normalizes top-level primitive values to the target parameter type before comparing:

- Resolves the target type per argument index, including `params` array element types, nullable underlying types, and default parameter values.
- Converts only when the conversion is lossless, verified by a round-trip check, so `0.1f` passed to a `double` parameter stays distinct from `0.1`.
- Applies the same normalization in `GetHashCode` so the hash set actually calls `Equals` for these pairs.
- Nested object arrays and non-numeric values keep their previous comparison semantics.


bradwilson:
Thanks!

PR #214, opened 2026-07-05
Stop reporting xUnit1010 for InlineData strings that convert to the parameter type at runtime

KitKeen:
Fixes xunit/xunit#2354

xUnit.net converts data row values to the test method parameter types via `Convert.ChangeType` (`Reflector.ConvertArgument` in v2, `TypeHelper.ConvertArgument` in v3), so `[InlineData("0.05")]` for a `decimal` parameter works at runtime but was reported as incompatible.

Following the suggestion in the issue, the supported conversion targets are now a lookup table (the from-string column of the `Convert` support table: `bool`, `char`, and the numeric types). Since InlineData values are compile-time constants, the analyzer checks the actual value: a string is only considered compatible when `Convert.ChangeType` to the target type succeeds under the invariant culture. Strings which fail to parse, such as `"Hello world"` for `int`, are still reported, since they fail the test at runtime.

The allowance is deliberately narrow, matching where the runtime actually converts:

- Only when the data row exactly fills the parameter list: `TypeHelper.ConvertArguments` skips conversion entirely on a length mismatch, so rows relying on default parameter values or params expansion keep strict checking.
- Not for params array elements, which are packed with `Array.Copy` without conversion.
- Not for `MemberData` arguments, which bind to the data method via reflection without conversion.
- Nullable and enum parameter types keep strict checking, since `Convert.ChangeType` does not handle them.

One caveat: the runtime converts with `CultureInfo.CurrentCulture`, while the analyzer parses with the invariant culture for deterministic analysis. A value like `"0.05"` could still fail at runtime on a machine whose culture uses a different decimal separator.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/X1010_InlineDataMustMatchTheoryParametersTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/X1010_InlineDataMustMatchTheoryParametersTests.cs
@@ -591,6 +591,101 @@ public class X1010_InlineDataMustMatchTheoryParametersTests
 				void ToGuid(Guid g) { }
 			}
 
+			// https://github.com/xunit/xunit/issues/2354
+			class NumericStringConversions {
+				[Theory]
+				[InlineData("42")]
+				[InlineData("-42")]
+				[InlineData({|xUnit1010:""|})]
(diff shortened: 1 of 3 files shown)


bradwilson:
Thanks!

PR #215, opened 2026-07-13
Add xUnit1070: Detect unused collection definitions

KitKeen:
Fixes xunit/xunit#1277

New rule xUnit1070 (Usage, Warning) flags a class marked with `[CollectionDefinition]` when nothing in the assembly references its collection, which catches misspelled collection names and definitions left behind after their tests were removed.

What counts as a reference:

- `[Collection(name)]` on any class, matched ordinally against the definition name.
- The v3 forms `[Collection(typeof(T))]` and `[Collection<T>]`, matched against the definition class itself. Unnamed v3 definitions participate this way and are reported by class name.
- The referencing attribute can sit on any named type, so definitions referenced only from abstract base classes or non-test classes are not flagged.

Analysis is skipped entirely when an assembly-level `[CollectionBehavior]` has constructor arguments: `CollectionPerAssembly` and custom collection factories attach definitions to collections without any `[Collection]` attribute, and flagging in that situation would be a false positive. `[assembly: CollectionBehavior(DisableTestParallelization = true)]` uses only named arguments and keeps the analysis active.

Since the answer is only known once the whole compilation has been seen, the diagnostic is reported from a compilation end action and the descriptor carries the `CompilationEnd` custom tag (the `Diagnostic` factory helper gained optional custom tags for this). As with other compilation-end diagnostics, it shows up on full builds rather than while typing.

The corollary discussed in the issue thread, a refactoring that generates a definition for a `[Collection]` name that has none, is not part of this PR since it needs to be an info-level analyzer with a fixer. Happy to tackle that as a follow-up.

### DIFF
--- a/src/xunit.analyzers/X1000/UnusedCollectionDefinitions.cs
+++ b/src/xunit.analyzers/X1000/UnusedCollectionDefinitions.cs
@@ -1,0 +1,90 @@
+using System;
+using System.Collections.Concurrent;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Xunit.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public class UnusedCollectionDefinitions : XunitDiagnosticAnalyzer
+{
+	public UnusedCollectionDefinitions() :
+		base(Descriptors.X1070_CollectionDefinitionIsNeverUsed)
+	{ }
+
+	public override void AnalyzeCompilation(
+		CompilationStartAnalysisContext context,
+		XunitContext xunitContext)
+	{
+		Guard.ArgumentNotNull(context);
+		Guard.ArgumentNotNull(xunitContext);
+
+		var collectionAttributeType = xunitContext.Core.CollectionAttributeType;
+		var collectionDefinitionAttributeType = xunitContext.Core.CollectionDefinitionAttributeType;
+		if (collectionAttributeType is null || collectionDefinitionAttributeType is null)
+			return;
+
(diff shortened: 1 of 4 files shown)


src/xunit.analyzers/X1000/UnusedCollectionDefinitions.cs
```diff
@@ -0,0 +1,90 @@
...
+	{
+		Guard.ArgumentNotNull(context);
+		Guard.ArgumentNotNull(xunitContext);
+
+		var collectionAttributeType = xunitContext.Core.CollectionAttributeType;
+		var collectionDefinitionAttributeType = xunitContext.Core.CollectionDefinitionAttributeType;
+		if (collectionAttributeType is null || collectionDefinitionAttributeType is null)
+			return;
+
+		var collectionBehaviorAttributeType = xunitContext.Core.CollectionBehaviorAttributeType;
```
bradwilson, line 28: I'm not sure I understand why `[CollectionBehavior]` has anything to do with this analyzer.

bradwilson:
I left a comment on the code, but I'll put a larger comment here, based on the body of the text:

> Analysis is skipped entirely when an assembly-level `[CollectionBehavior]` has constructor arguments: `CollectionPerAssembly` and custom collection factories attach definitions to collections without any `[Collection]` attribute, and flagging in that situation would be a false positive. `[assembly: CollectionBehavior(DisableTestParallelization = true)]` uses only named arguments and keeps the analysis active.

I don't know why this decision was made, and I'm not sure I agree with it. Regardless of what the default behavior is, placing `[Collection]` and `[CollectionDefinition]` overrides that default behavior, and thus should still be correctly specified. A collection definition with no consumers is still incorrect behavior.

bradwilson:
This PR got closed when the `rel/2.0.0` branch was merged into `main`. Please re-open the PR against `main`. Thanks!

PR #216, opened 2026-07-16
Add xUnit2033: Use the assertion return value instead of re-deriving it

KitKeen:
Fixes xunit/xunit#1506

## Summary

`Assert.Single`, `Assert.IsType<T>`, and `Assert.IsAssignableFrom<T>` return useful values (the single item / the typed value), but tests often throw the value away and then re-derive it:

```csharp
Assert.Single(filters);
Filter filter = filters.Single();

Assert.IsType<MyType>(value);
var typed = (MyType)value;
```

This PR adds **xUnit2033: Use the assertion return value instead of re-deriving it** (Info severity, Assertions category), plus a code fixer that turns the assertion into `var item = Assert.Single(filters);` and replaces the later re-derivation with the new variable.

## Design: only provable re-derivations are flagged

To keep the rule noise-free, the diagnostic is reported only when both hold:

- the assertion's return value is discarded (the call is a standalone expression statement), and
- a later statement in the same block re-derives the same value:
  - after `Assert.Single(xs)`: a parameterless `xs.Single()` / `xs.First()` / `xs.ElementAt(0)` / `xs[0]`
  - after `Assert.IsType<T>(x)` / `Assert.IsAssignableFrom<T>(x)`: `(T)x` or `x as T` (same `T` by symbol comparison)
  - where `xs` / `x` are compared as syntactically equivalent expressions

Deliberately not flagged:

- the guard idiom (`Assert.Single(xs);` with no later re-derivation): discarding the return value on purpose is legitimate and common, so the plain "return value unused" case stays silent
- the `Single(collection, expected)` and `Single(collection, predicate)` overloads (they assert something different), and predicate-based `xs.Single(...)` / `xs.First(...)` re-derivation candidates
- anything after the root identifier may have changed (`xs = ...`, `ref` / `out` argument, increment / decrement)
- matches inside lambdas or local functions (they run at another time)

The fixer picks a collision-safe variable name (`item` / `typed`, suffixed with `_2` etc. when taken), mirroring the xUnit2023 fixer.

## Tests

TDD: analyzer tests were written first against a no-op skeleton (RED 3/6) and driven to GREEN; fixer tests likewise (RED 3/3, then GREEN). Full suite: net8.0 304/304, net472 291/291.

### DIFF
--- a/src/xunit.analyzers.fixes/X2000/AssertReturnValueShouldBeUsedFixer.cs
+++ b/src/xunit.analyzers.fixes/X2000/AssertReturnValueShouldBeUsedFixer.cs
@@ -1,0 +1,133 @@
+using System.Collections.Immutable;
+using System.Composition;
+using System.Globalization;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Text;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Xunit.Analyzers.Fixes;
+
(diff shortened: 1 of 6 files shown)


bradwilson:
Thanks!

PR #217, opened 2026-09-03
xunit/xunit#3620: Add Fix All support for xUnit2033

### LINKED ISSUE #3620: [FR] Add Fix All support for xUnit2033
### Description

Currently, the code fix for **xUnit2033** can only be applied to one diagnostic at a time in Visual Studio 2022.

It would be useful if the analyzer provided **Fix All** support, allowing the code fix to be applied at the following scopes:

* Document
* Project
* Solution

### Current behavior

When xUnit2033 is reported multiple times, Visual Studio only offers the fix for the individual occurrence.

This requires manually applying the same code fix to every diagnostic.

### Expected behavior

The code fix should support Visual Studio's standard **Fix All** functionality:

* Fix all occurrences in the current document
* Fix all occurrences in the current project
* Fix all occurrences in the entire solution

### Motivation

xUnit2033 can appear many times when migrating or cleaning up an existing test suite.

Since the transformation is already available as an individual code fix, supporting **Fix All** would make it significantly easier to apply the same change consistently across larger projects and solutions.

rghvgrv:
Fixes xunit/xunit#3620.

`AssertReturnValueShouldBeUsedFixer` inherited the default `GetFixAllProvider()` from `XunitCodeFixProvider`, which returns `null`, so the xUnit2033 code fix could only be applied one diagnostic at a time. Every other X2000 fixer already overrides it with `BatchFixer`.

Changes:
- `AssertReturnValueShouldBeUsedFixer` now overrides `GetFixAllProvider()` with `WellKnownFixAllProviders.BatchFixer`, matching the surrounding fixers. The code action already supplies the `Key_UseReturnValue` equivalence key, so no other change was needed to make batching work. This enables Fix All at document, project, and solution scope.
- The existing fixer test switches from `Verify.VerifyCodeFix` to `Verify.VerifyCodeFixFixAll`, the convention used by the fixer tests whose fixers support Fix All. It already contained three diagnostics in a single document, including the safe-variable-name collision case, so it exercises the batched path directly.

Tests pass against xUnit.net v2 and v3.


rghvgrv:
> @rghvgrv please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

bradwilson:
This requires more than just turning on the flag, because it generates identifiers in the fixed source code.

If you change the first example to:

```csharp
[Fact]
public void UsesSingleReturnValue() {
	var xs = new List<int> { 42 };
	[|Assert.Single(xs)|];
	var value = xs.Single();
	[|Assert.Single(xs)|];
	var value2 = xs.Single();
}
```

Then the fixed source should be:

```csharp
[Fact]
public void UsesSingleReturnValue() {
	var xs = new List<int> { 42 };
	var item = Assert.Single(xs);
	var value = item;
	var item_2 = Assert.Single(xs);
	var value2 = item_2;
}
```

This is what you get today without batch fixing. When you turn the batch fixing on as you did, then the generated source is no longer legal (because of the doubly-declared `item` variable):

```csharp
[Fact]
public void UsesSingleReturnValue() {
	var xs = new List<int> { 42 };
	var item = Assert.Single(xs);
	var value = item;
	var item = Assert.Single(xs);
	var value2 = item_2;
}
```

I've just pushed (to `main`) an updated test for xUnit2033's fixer to illustrate this need. I'm also removing the FixAll verification functions, because they're identical in behavior to the non-FixAll versions, which means the fix should not need to touch the test at all, just make it work when enabling batching (it will fail with the new test). I'm push an update to your PR branch to merge in `main` to pick up the new test (and remove the change to account for my removal of `FixAll`).

Thanks!

PR #218, opened 2026-09-12
xUnit1041: Fix crash on [assembly: AssemblyFixture] without an argument

meziantou:
## Summary

`EnsureFixturesHaveASource` (xUnit1041) read `ConstructorArguments[0]` for every `AssemblyFixture` attribute without checking that an argument exists. When the attribute has no argument (which happens all the time while typing), the analyzer threw `IndexOutOfRangeException` (AD0001).

Assembly attributes are checked for every test class, so one incomplete `[assembly: AssemblyFixture]` crashed the analyzer for every test class and turned off xUnit1041 across the whole project.

## Repro

```csharp
using Xunit;

[assembly: AssemblyFixture]

public class TestClass {
    public TestClass(int x) { }

    [Fact]
    public void TestMethod() { }
}
```

- **Expected:** xUnit1041 on `x`
- **Actual:** `AD0001: Analyzer 'Xunit.Analyzers.EnsureFixturesHaveASource' threw an exception of type 'System.IndexOutOfRangeException'`

## Changes

- Skip `AssemblyFixture` attributes that have no constructor arguments. This uses the same `IsDefaultOrEmpty` check the analyzer already uses for `CollectionDefinition`.
- Add the regression test `V3_AssemblyFixtureWithoutArgument_DoesNotCrash`, which uses the repro above with `CompilerDiagnostics.None`.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/X1041_EnsureFixturesHaveASourceTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/X1041_EnsureFixturesHaveASourceTests.cs
@@ -1,5 +1,6 @@
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Testing;
 using Xunit;
 using Verify = CSharpVerifier<Xunit.Analyzers.EnsureFixturesHaveASource>;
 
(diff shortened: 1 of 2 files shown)


bradwilson:
Since you're doing many PRs, just as an FYI, I've added this note to BUILDING about adding tests:

https://github.com/xunit/xunit.analyzers/blob/main/BUILDING.md#adding-tests

> 

It's my intention to review and update your tests as appropriate before merging the PRs, so there's nothing to do here right now. However, moving forward it would be most excellent (read: less effort for me 😆) if you just added test scenarios to existing tests instead of writing new tests whenever possible.

bradwilson:
Thanks!

bradwilson:
Fix available in `2.2.0-pre.3` https://xunit.net/docs/using-ci-builds

PR #219, opened 2026-09-12
Fix xUnit1044/xUnit1045 not analyzing MemberData with MemberType

meziantou:
## Problem

xUnit1044 and xUnit1045 never analyzed a `MemberData` attribute that points at another class through `MemberType`. This affected both v2 and v3.

`GetMemberContainingType` read `namedArgument.Value.Type`, which is the type of the `typeof(...)` constant (`System.Type`), not the type it names. So the analyzer looked up the member on `System.Type`, didn't find it, and reported nothing.

```csharp
public sealed class NonSerializable { }

public class DataSource {
    public static TheoryData<NonSerializable> Data => new TheoryData<NonSerializable>();
}

public class TestClass {
    [Theory]
    [MemberData(nameof(DataSource.Data), MemberType = typeof(DataSource))] // expected xUnit1044, got nothing
    public void Test(NonSerializable p) { }
}
```

## Fix

The analyzer now uses the constant's value instead: `namedArgument.Value.Value as ITypeSymbol`.

### DIFF
--- a/src/xunit.analyzers.tests/Analyzers/X1000/X1044_TheoryDataTypeArgumentsShouldBeSerializableTests.cs
+++ b/src/xunit.analyzers.tests/Analyzers/X1000/X1044_TheoryDataTypeArgumentsShouldBeSerializableTests.cs
@@ -554,6 +554,43 @@ public class X1044_TheoryDataTypeArgumentsShouldBeSerializableTests
 	}
 
 	[Fact]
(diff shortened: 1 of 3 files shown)


bradwilson:
Thanks!

bradwilson:
Fix available in `2.2.0-pre.4` https://xunit.net/docs/using-ci-builds

PR #220, opened 2026-09-12
Fix xUnit1007 analyzer crash on [ClassData()] with no arguments

meziantou:
## What

`ClassDataAttributeMustPointAtValidClass` (xUnit1007, 1037–1040, 1050) now skips `[ClassData]` attributes whose argument list is empty.

## Why

The analyzer checked `ArgumentList` for null, then read `ArgumentList.Arguments[0]` without checking the count. For `[ClassData()]` this threw an `ArgumentOutOfRangeException`, reported as AD0001 on top of the compiler error:

```csharp
public class TestClass {
    [Theory]
    [ClassData()]
    public void TestMethod(int n) { }
}
```

The code already fails to compile (CS7036), but the analyzer crash is noisy in the IDE while the attribute is still being typed.


bradwilson:
Thanks!

bradwilson:
Fix available in `2.2.0-pre.5` https://xunit.net/docs/using-ci-builds

PR #221, opened 2026-09-12
Forward CancellationToken to Roslyn APIs in rules

meziantou:
Several analyzers, suppressors, and code fixers called Roslyn APIs such as `GetTypeInfo`, `GetSymbolInfo`, `GetDeclaredSymbol`, `GetSyntax`, `GetSyntaxAsync` and `GetRoot` without passing the available cancellation token. This means analysis couldn't be cancelled promptly, for example while the user is typing in the IDE.

## Changes

- **Direct forwarding (28 call sites):** these calls now pass `context.CancellationToken`.
- **Helpers that had no token (6 call sites):** these methods now take a required `CancellationToken` parameter, and callers forward theirs:
  - `CodeAnalysisExtensions.IsInTestMethod` and `IsPointer`
  - `SyntaxExtensions.ContainsAttributeType` (the `SyntaxList<AttributeListSyntax>` overload)
  - `BooleanAssertsShouldNotBeUsedForSimpleEqualityCheck.LiteralReferenceKind`
  - `DoNotUseBlockingTaskOperations`: `FindSymbol`, `TaskIsKnownToBeCompleted`, `ValidateTasksInWhenAll`, `ValidateTaskFromWhenAny`
  - `AssertReturnValueShouldBeUsed`: `FindRederivation`, `GetTypedRederivation`
  - `MemberDataShouldReferenceValidMember.GetParameterExpressionsFromArrayArgument`
  - `TestMethodWithTimeoutShouldUseCancellationToken.GetTimeoutLocation`
  - `InlineDataMustMatchTheoryParameters_TooFewValuesFixer.CreateDefaultValueSyntax`

## Notes for reviewers

- The token parameters are required rather than optional, so new call sites can't silently drop the token. This changes the signatures of the public utility methods listed above. Nothing outside these projects calls them.
- To find these call sites, I temporarily added [Meziantou.Analyzer](https://github.com/meziantou/Meziantou.Analyzer) with only MA0032 and MA0040 enabled. It isn't part of this PR. With the changes applied, it reports no warnings.
- The net8.0 tests pass for both `xunit.analyzers.tests` and `xunit.analyzers.latest.tests`. I didn't run the net472 tests.

### DIFF
--- a/src/xunit.analyzers.fixes/X1000/InlineDataMustMatchTheoryParameters_TooFewValuesFixer.cs
+++ b/src/xunit.analyzers.fixes/X1000/InlineDataMustMatchTheoryParameters_TooFewValuesFixer.cs
@@ -73,7 +73,7 @@ public class InlineDataMustMatchTheoryParameters_TooFewValuesFixer : XunitCodeFi
 		var originalInitializer = arrayInitializer;
 		var i = originalInitializer?.Expressions.Count ?? attribute.ArgumentList?.Arguments.Count ?? 0;
 		for (; i < method.ParameterList.Parameters.Count; i++)
-			if (CreateDefaultValueSyntax(editor, method.ParameterList.Parameters[i].Type) is ExpressionSyntax defaultExpression)
+			if (CreateDefaultValueSyntax(editor, method.ParameterList.Parameters[i].Type, cancellationToken) is ExpressionSyntax defaultExpression)
 			{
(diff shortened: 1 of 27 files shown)


bradwilson:
Thanks!

bradwilson:
Fix available in `2.2.0-pre.6` https://xunit.net/docs/using-ci-builds

PR #222, opened 2026-09-13
Convert xUnit1054/xUnit1055 analyzer to IOperation API

meziantou:
Converts `ConditionalSkipPropertyValidation` (xUnit1054 / xUnit1055) from a syntax node action to the IOperation API.

## What changed

- The attribute type comes from the `IObjectCreationOperation` wrapped by `IAttributeOperation`, rather than a `SemanticModel.GetTypeInfo` call.
- Named arguments are read from the object initializer's `ISimpleAssignmentOperation`s, matching on the target property instead of the `NameEquals` text.
- `SkipType` is read from `ITypeOfOperation.TypeOperand`.
- `SkipUnless` / `SkipWhen` use the argument's `ConstantValue`. This replaces the old `InvocationExpressionSyntax` branch that split the `nameof(...)` argument text on `.` (and resolves its TODO).
- The property lookup uses `GetMembers(propertyName)` instead of filtering all members.

## Notes for reviewers

The old invocation branch only existed to handle `nameof(...)`, which the parser sees as an invocation. As operations it is an `INameOfOperation` with a constant value. I checked how the argument forms behave:

| Argument | Operation | Constant |
|---|---|---|
| `nameof(P)` / `nameof(C.P)` | NameOf | `P` |
| `nameof(List<int>)` | NameOf | `List` (old code produced `List<int>`) |
| `const string K = "P"` | FieldReference | `P` (previously ignored) |
| `"P"` | Literal | `P` |
| `nameof(DoesNotExist)` | NameOf | `DoesNotExist` (still reported) |
| `M("P")` | Invocation | none (already a CS0182 compile error) |

Each assignment's syntax is the `AttributeArgumentSyntax`, so diagnostic locations are unchanged.

BTW, would you be interested in more PRs to convert syntax API to Symbol or Operation when possible?


bradwilson:
> BTW, would you be interested in more PRs to convert syntax API to Symbol or Operation when possible?

I assume the operation versions have better performance? Or is it a matter of code simplification?

meziantou:
Performance: it depends what each rule is doing. Querying the semantic model from the syntax has a cost, using Operation is often better as Roslyn compute stuff once to create the operation, then call the handler when it's ready.

The main benefits are correctness and simplification: dealing with syntax is complex and it's very easy to forget cases and have broken rules because of new C# syntaxes.

For instance, in this rule the code is using `InvocationExpressionSyntax` to try to find the name. Instead, it should have ask for the constant value using the semantic model to handle `nameof` and other valid syntaxes. Using Operation I think it's easier to use the right thing.

meziantou:
Also, dealing with syntax means you have to be way more defensive. The syntax may be invalid (most of the time), so lots of null check and validation. Operation removes most of those checks.

bradwilson:
Excellent summary, thanks! I assume then that I will try to use operations whenever possible, and if you are so motivated to do the conversions of existing code, I'll happily accept the PRs. 😄

bradwilson:
Thanks!

bradwilson:
Fix available in `2.2.0-pre.7` https://xunit.net/docs/using-ci-builds

PR #223, opened 2026-09-13
Rewrite xUnit1029 (LocalFunctionsCannotBeTestFunctions) using IOperation

meziantou:
This PR rewrites `LocalFunctionsCannotBeTestFunctions` (xUnit1029) to use the `IOperation` API instead of syntax nodes.

## Changes
- The analyzer now registers for `OperationKind.LocalFunction` instead of `SyntaxKind.LocalFunctionStatement`.
- It reads attributes from `ILocalFunctionOperation.Symbol.GetAttributes()` and matches `AttributeClass` against the Fact/Theory and data attribute types. It no longer calls `SemanticModel.GetSymbolInfo` on each attribute.
- The list of attribute types is built once per compilation instead of once per local function. If none of those types exist, the analyzer returns without registering an action.
- The diagnostic location and the `[...]` message argument still come from the attribute's source text, reached through `AttributeData.ApplicationSyntaxReference`. This keeps the message the same as before.

## Notes for reviewers
- Behavior change: the old code skipped an attribute whose constructor didn't resolve, such as one with wrong arguments. The new code reports it as long as the attribute type resolves, since the local function is still marked as a test either way.
- The existing `X1029_LocalFunctionsCannotBeTestFunctionsTests` pass unchanged. So does the rest of the net8.0 test run. I didn't run the net472 tests locally.


bradwilson:
(Waiting to merge to understand the seeming duplication with #226)

meziantou:
Closing in favor of #226

PR #224, opened 2026-09-13
Convert xUnit1004 analyzer to IOperation API

meziantou:
Converts `TestMethodShouldNotBeSkipped` (xUnit1004) from a syntax node action to the IOperation API, following the same approach as #222.

## What changed

- Named arguments are read from the object initializer's `ISimpleAssignmentOperation`s, matching on the target property name instead of the `NameEquals` text.
- The attribute type comes from the `IObjectCreationOperation` wrapped by `IAttributeOperation`, rather than a `SemanticModel.GetTypeInfo` call. The `IsAssignableFrom` check is kept, so attributes derived from `FactAttribute` / `TheoryAttribute` are still covered.
- Added `Constants.AttributeProperties.Skip` alongside the existing `SkipUnless` / `SkipWhen` constants.

## Notes for reviewers

- Behavior is unchanged: the diagnostic is suppressed when `SkipUnless` or `SkipWhen` is set, and is reported on the `Skip = ...` argument. Each assignment's syntax is the `AttributeArgumentSyntax`, so diagnostic locations are unchanged and the fixer (which looks for `AttributeArgumentSyntax`) needs no update.
- The existing `V2_and_V3` test is extended with an attribute derived from `FactAttribute` (with and without `Skip`) and a `Skip` value coming from a `const`.


bradwilson:
Looks like the tests failed because of the addition of the test cases deriving from `FactAttribute`, which is not legal in Native AOT.

meziantou:
The CI should be green now

bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.20` https://xunit.net/docs/using-ci-builds

PR #225, opened 2026-09-13
Convert xUnit1058 analyzer to symbol API

meziantou:
Converts `GenericCollectionDefinitionsInAot` (xUnit1058) from a syntax node action to the symbol API.

## What changed

- Registers a `SymbolKind.NamedType` action instead of a `SyntaxKind.Attribute` action; the `Microsoft.CodeAnalysis.CSharp` / `.Syntax` usings are gone.
- The `[CollectionDefinition]` attribute is found through `INamedTypeSymbol.GetAttributes()` rather than resolving each `AttributeSyntax` with `SemanticModel.GetTypeInfo` and then using `ContainingSymbol`.
- The generic check (`IsGenericType`), exact attribute type match, and diagnostic location (`Locations.FirstOrDefault()`) are unchanged.

## Notes for reviewers

- Because the analysis now runs once per type, a type is reported at most once, even if its declaration were somehow split across attribute occurrences.
- No test changes: the existing `X1058_GenericCollectionDefinitionsInAotTests` passes in both `xunit.analyzers.tests` and `xunit.analyzers.latest.tests` (the AOT assertion only compiles under `ROSLYN_LATEST`). I also confirmed the test fails if the generic check is inverted, so it exercises the new code path.


bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.8` https://xunit.net/docs/using-ci-builds

PR #226, opened 2026-09-13
Convert xUnit1029 analyzer to IOperation API

meziantou:
Converts `LocalFunctionsCannotBeTestFunctions` (xUnit1029) from a syntax node action to the IOperation / symbol APIs.

## What changed

- Registers an `OperationKind.LocalFunction` action instead of a `SyntaxKind.LocalFunctionStatement` action; the `Microsoft.CodeAnalysis.CSharp` / `.Syntax` usings are gone.
- Attributes come from `ILocalFunctionOperation.Symbol.GetAttributes()`, checking `AttributeData.AttributeClass`, rather than calling `SemanticModel.GetSymbolInfo` on each `AttributeSyntax`.
- The Fact/Theory/Data attribute type list is built once per compilation instead of on every local function. If none are available, no action is registered.

## Notes for reviewers

- Syntax is still used for the diagnostic location and message text (`[Fact]`, `[InlineData(42)]`, …), via `AttributeData.ApplicationSyntaxReference`. Locations and messages are unchanged, so the fixer still works as before.
- I added scenarios to the existing `V2_and_V3` test: a local function inside a lambda, and a nested local function with two attributes in one list.
- Tested on net8.0 for both `xunit.analyzers.tests` and `xunit.analyzers.latest.tests` (analyzer + fixer tests). net472 was not run (macOS).


bradwilson:
How does this differ from #223 other than the additional tests?

bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.19` https://xunit.net/docs/using-ci-builds

PR #227, opened 2026-09-13
Convert xUnit1060 analyzer to IOperation API

meziantou:
Converts `CulturedTestMustHaveAtLeastOneCulture` (xUnit1060) from a syntax node action to the IOperation API.

## What changed

- Registers an `OperationKind.Attribute` action instead of a `SyntaxKind.Attribute` action. The `Microsoft.CodeAnalysis.CSharp` and `.Syntax` usings are gone.
- The attribute type and the `cultures` argument come from the `IObjectCreationOperation` wrapped by `IAttributeOperation`, instead of a `SemanticModel.GetTypeInfo` call and `ArgumentList.Arguments[0]`.
- Implicit conversions are unwrapped from the argument first, then:
  - `IArrayCreationOperation`: reported when the dimension size is the constant `0`. This covers `new string[0]` and `new string[] { }`, because an initializer sets the size.
  - `ICollectionExpressionOperation`: reported when `Elements.IsEmpty`.
- The diagnostic is still reported on the whole attribute.

## Notes for reviewers

- **Behavior change:** the old code reported any `ArrayCreationExpressionSyntax` without an initializer, so `[CulturedFact(new string[3])]` (three `null` cultures) was flagged. It isn't flagged anymore, because the array isn't empty. I can restore the old behavior if you'd rather keep it.
- Both the `ICollectionExpressionOperation` branch and the conversion unwrap are needed. In an attribute argument, `[]` shows up wrapped in an implicit conversion to `string[]`. With either one removed, the three `[]` cases in the tests stop reporting.
- Added test scenarios to the existing test: `new string[] { "en-US", "fr-FR" }` and `cultures: ["en-US"]` (no diagnostic), plus `cultures: []` and `cultures: new string[] { }` (diagnostic).


bradwilson:
> * **Behavior change:** the old code reported any `ArrayCreationExpressionSyntax` without an initializer, so `[CulturedFact(new string[3])]` (three `null` cultures) was flagged. It isn't flagged anymore, because the array isn't empty. I can restore the old behavior if you'd rather keep it.

This detection (illegal and/or duplicate cultures) should be handled by a separate analyzer (even if they share the base analysis code). I'll open a new issue for it.

bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.9` https://xunit.net/docs/using-ci-builds

PR #228, opened 2026-09-14
Convert xUnit1057 analyzer to IOperation API

meziantou:
Converts `TypeMustBePublicOrInternal` (xUnit1057) from a syntax node action to the IOperation API, following the same approach as #222.

## What changed

- The attribute type comes from the `IObjectCreationOperation` wrapped by `IAttributeOperation`, rather than a `SemanticModel.GetTypeInfo` call.
- `SkipExceptions` is read from the object initializer's `ISimpleAssignmentOperation`, matching on the target property instead of the `NameEquals` text.
- The three separate syntax branches (`new Type[] { … }`, `new[] { … }`, `[ … ]`) become one switch over `IArrayCreationOperation` and `ICollectionExpressionOperation`, and exception types come from `ITypeOfOperation.TypeOperand`.
- The fixture check (`IClassFixture<>` / `ICollectionFixture<>`) was already symbol-based and is unchanged.

## Notes for reviewers

- Diagnostic locations are unchanged: the attribute operation's syntax is the `AttributeSyntax`, and each `ITypeOfOperation`'s syntax is the `typeof(...)` expression. Note that switching to ISymbol would make the code simpler, however it would not report on the `typeof` expression but the full attribute.


bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.10` https://xunit.net/docs/using-ci-builds

PR #229, opened 2026-09-14
Convert xUnit1007/xUnit1037-1040/xUnit1050 analyzer to IOperation API

meziantou:
Converts `ClassDataAttributeMustPointAtValidClass` (xUnit1007, xUnit1037, xUnit1038, xUnit1039, xUnit1040, xUnit1050) from a syntax node action to the IOperation API.

## What changed

- Registers an `OperationKind.Attribute` action instead of a `SyntaxKind.MethodDeclaration` action that walked every attribute list.
- The test method is `context.ContainingSymbol`, so `SemanticModel.GetDeclaredSymbol` is no longer needed.
- The attribute type comes from the `IObjectCreationOperation` wrapped by `IAttributeOperation`, rather than `SemanticModel.GetTypeInfo`.
- `[ClassData(typeof(X))]`: the class type is `ITypeOfOperation.TypeOperand` from the first constructor argument. A missing argument (the #220 case) just doesn't match.
- `[ClassData<X>]`: compares `attributeType.OriginalDefinition` with `ClassDataAttributeOfTType` once, instead of constructing unbound generic types for every generic attribute.
- Parameters come only from `IMethodSymbol.Parameters`; the parallel `ParameterList` syntax walk is gone.
- The report helpers take a `Location` instead of `AttributeSyntax` / `TypeSyntax`.

## Notes for reviewers

- xUnit1039/xUnit1040 are reported on the parameter's *type*. The symbol only exposes the parameter name location, so a small helper still reads `ParameterSyntax.Type` from `DeclaringSyntaxReferences` to keep those locations the same.


bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.11` https://xunit.net/docs/using-ci-builds

PR #230, opened 2026-09-14
xUnit2025: Don't suggest simplification for bool? when it changes null behavior

meziantou:
## Problem

xUnit2025 suggests simplifying boolean comparisons into `Assert.True`/`Assert.False` even when the operand is `bool?`. `Assert.True(bool?)` and `Assert.False(bool?)` both **fail** on `null`, but `null != true` evaluates to `true`, so some of these rewrites turn a passing assertion into a failing one:

```csharp
void M(bool? nb) {
    Assert.True(nb != true);    // was fixed to Assert.False(nb)
    Assert.False(nb == true);   // was fixed to Assert.False(nb)
    Assert.True(nb != false);   // was fixed to Assert.True(nb)
}
```

When `nb` is `null`, each original assertion passes but its replacement throws.

## Fix

When the non-literal operand is `Nullable<bool>`, the diagnostic is now skipped only for the forms where the original assertion passes on `null`:

| Form | `null`: original | `null`: replacement | Diagnostic |
|---|---|---|---|
| `Assert.True(nb == X)` | fails | fails | kept |
| `Assert.False(nb != X)` | fails | fails | kept |
| `Assert.True(nb != X)` | passes | fails | **skipped** |
| `Assert.False(nb == X)` | passes | fails | **skipped** |

The rewrites that behave the same on `null` (for example `Assert.True(nb == true)` → `Assert.True(nb)`) are still reported and fixed.

## Tests

The `bool?` scenarios are added to the existing `V2_and_V3` tests rather than new test methods:
- **Analyzer test:** all 16 combinations (both asserts, both operators, both literals, literal on either side, with and without a message).
- **Fixer test:** the `bool?` forms that still get fixed produce the right replacement, and the skipped forms are left alone.


bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.12` https://xunit.net/docs/using-ci-builds

PR #231, opened 2026-09-14
Fix xUnit1013 false positive on compiler-generated record Deconstruct

meziantou:
## What

xUnit1013 now skips compiler-generated methods (`IMethodSymbol.IsImplicitlyDeclared`).

## Why

A positional record gets a compiler-generated `public void Deconstruct(out ...)`. It is public, returns void and isn't a test, so xUnit1013 reported it on the record's name. The user can't remove that method or put an attribute on it.

```csharp
using Xunit;

public record TestClass(int Value) {
    [Fact]
    public void TestMethod() { }
}
```

Before this change, both v2 and v3 reported `xUnit1013: Public method 'Deconstruct' on test class 'TestClass' should be marked as a Theory`.

## Tests

- I added a positional record scenario to the existing `V2_and_V3` test. The record also has a hand-written public method, and the test expects xUnit1013 to still report it.
- Records need C# 9, so that test now compiles with `LanguageVersion.CSharp9` instead of the default C# 6. None of its other scenarios depend on C# 6. The source also declares `IsExternalInit`, so records compile against the v2 reference assemblies.
- Without the new check, `V2_and_V3` fails only because of the extra `Deconstruct` diagnostic. With the check, all tests pass on net8.0.


bradwilson:
I don't mind this fix, since it makes sense regardless of why there is an implicitly defined method. I am curious what the use case is for record-based tests, though. 😄

bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.13` https://xunit.net/docs/using-ci-builds

PR #232, opened 2026-09-14
MemberData analyzer: analyze test methods declared in records

meziantou:
## Summary

All MemberData rules (xUnit1014–1021, 1036–1042, 1053, …) were silently skipped for test methods declared inside a `record`, `record class`, or `record struct`.

`MemberDataShouldReferenceValidMember.GetClassTypesForAttribute` found the containing type with `FirstAncestorOrSelf<ClassDeclarationSyntax>()`. Records are `RecordDeclarationSyntax`, which does not derive from `ClassDeclarationSyntax`, so the lookup returned `null` and the analyzer returned early.

### Repro

```csharp
using Xunit;

public record TestClass {
    [Theory]
    [MemberData("Missing")]
    public void Test(int _) { }
}
```

- **Expected:** xUnit1015
- **Before this PR:** no diagnostics

## Fix

The lookup now uses `FirstAncestorOrSelf<TypeDeclarationSyntax>()`, which matches classes, records, and structs. The two code fixers that call this helper (`ExtraValueFixer`, `NullShouldNotBeUsedForIncompatibleParameterFixer`) get the fix too.

## Tests

- Added record, record class, and record struct scenarios to the existing `X1015_MemberDataShouldReferenceValidMemberTests.V2_and_V3` test, and raised its language version from C# 8 to C# 10 so `record struct` compiles.
- Checked that the new scenarios fail without the fix (0 diagnostics returned, 3 expected).
- `xunit.analyzers.tests` and `xunit.analyzers.latest.tests` pass on net8.0. The net472 tests were **not** run because this was tested on macOS.


bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.14` https://xunit.net/docs/using-ci-builds

PR #233, opened 2026-09-14
Consider inherited test methods when detecting test classes (CA1515 suppressor, xUnit1056)

meziantou:
## Problem

`IsTestClassNonStrict` only inspected `type.GetMembers()`, ignoring base types. A concrete class whose test methods are all inherited was not recognized as a test class:

```csharp
using Xunit;

public abstract class BaseTests {
    [Fact]
    public void TestMethod() { }
}

public class DerivedTests : BaseTests { }
```

The CA1515 suppressor (`MakeTypesInternalSuppressor`) suppressed the warning on `BaseTests` but still reported it on `DerivedTests` — the class xUnit actually runs, which must be public.

## Fix

`IsTestClassNonStrict` now uses `GetInheritedAndOwnMembers()`, so inherited `[Fact]`/`[Theory]` methods are taken into account. The abstract base class remains suppressed as well (it cannot be made internal while a public test class derives from it — CS0060).

## Reviewer notes

The non-strict check is also used by **xUnit1056** (`TypeMustHaveSinglePublicConstructor`). With this change, a class that inherits its tests and declares multiple public constructors is now flagged, which is correct since xUnit cannot construct it.

## Tests

- CA1515 suppressor: added an abstract base test class and a derived class with inherited tests (both suppressed).
- xUnit1056: added a derived class with a default constructor (no diagnostic) and one with two public constructors (diagnostic).

Both new scenarios fail without the fix; the full test suite passes on net8.0.


bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.15` https://xunit.net/docs/using-ci-builds

PR #234, opened 2026-09-14
Continue analyzing remaining MemberData attributes after reporting an error

meziantou:
## Problem

`MemberDataShouldReferenceValidMember` loops over every `[MemberData]` attribute on a test method, but several error paths used `return` inside that loop instead of `continue`. After one attribute produced a diagnostic, the analyzer skipped all the remaining `[MemberData]` attributes on the method.

```csharp
public class TestClass {
    public TheoryData<int> NonStatic => null;

    [Theory]
    [MemberData("Missing")]          // xUnit1015 reported
    [MemberData(nameof(NonStatic))]  // xUnit1017 was never reported
    public void Test(int _) { }
}
```

## Fix

The five early exits in the attribute loop now `continue` instead of `return`:
- xUnit1057: member type must be public or internal (AOT)
- xUnit1068: member type is an open generic (AOT)
- xUnit1015: member doesn't exist
- xUnit1065: method is ambiguous
- xUnit1018: member isn't a field, method or property

## Tests

I added the repro to the existing `X1015_MemberDataShouldReferenceValidMemberTests.V2_and_V3` test instead of creating a new test method. Without the fix, the test fails with `expected "2" actual "1"`. All net8.0 tests pass.


bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.16` https://xunit.net/docs/using-ci-builds

PR #235, opened 2026-09-14
Add Fix All support for xUnit2033

### LINKED ISSUE #3620: [FR] Add Fix All support for xUnit2033

meziantou:
Fixes xunit/xunit#3620.

The xUnit2033 code fix could only be applied one diagnostic at a time. #217 tried to enable `WellKnownFixAllProviders.BatchFixer`, but as noted there, the fix introduces new local variables (`item`, `typed`), and the batch fixer computes each fix independently, producing duplicate declarations (e.g. two `var item = ...` in the same method).

## Changes

`AssertReturnValueShouldBeUsedFixer` now returns a document-based provider (`FixAllProvider.Create`) that applies all fixes in a document in a single pass, in source order:

- **Unique names**: it tracks the variables it has already introduced, so a second fix in the same (or an enclosing/nested) scope gets `item_2`. Sibling scopes (e.g. `if`/`else` branches) can still reuse `item`.
- **Same result as fixing one at a time**: when an earlier fix's re-derivation overlaps a later diagnostic's assertion or re-derivation, that later diagnostic would no longer be reported after the earlier fix, so it is skipped. Examples: two asserts sharing a single `xs.Single()`, or `xs.Single()` used as the argument of a later `Assert.IsType<int>(...)`.
- **Nested edits**: node replacements use the callback overloads of `DocumentEditor.ReplaceNode` so a re-derivation replaced inside a later assertion composes with that assertion's rewrite.
- The single-diagnostic code path shares the same helpers; its behavior is unchanged.

## Tests

Added scenarios to the existing `X2033_AssertReturnValueShouldBeUsedFixerTests` (the verifier automatically checks Fix All in document/project/solution when a provider exists):

- name reuse in sibling scopes
- two assertions sharing one re-derivation
- a re-derivation inside a later assertion

The existing test (from `main`) already covers the duplicate-name case. I confirmed the test fails on "Fix all in document" when switching the provider to `BatchFixer`, and passes with the new provider, against both xUnit.net v2 and v3, and in `xunit.analyzers.latest.tests`.

## Known limitation (pre-existing, not addressed here)

If an earlier fix declares `item` in a nested block and a later assertion is in the enclosing block, `LookupSymbols` at the later position doesn't see the nested local, so the name collides (CS0136). This already happens when applying fixes one at a time.


bradwilson:
Thanks!

bradwilson:
Available in `2.0.0-pre.17` https://xunit.net/docs/using-ci-builds

PR #236, opened 2026-09-19
Add xUnit1070/xUnit1071: cultured tests cannot have null or duplicate cultures

### LINKED ISSUE #3632: Add analyzer to detect illegal and/or duplicate cultures
For `[CulturedFact]` and `[CulturedTheory]`, we should add an analyzer which detects illegal (`null`) cultures, as well as duplicated cultures (e.g., `[CulturedFact(cultures: ["en-US", "en-US"])]`).

https://github.com/xunit/xunit.analyzers/pull/227#issuecomment-5701563325

meziantou:
Fixes xunit/xunit#3632.

Adds two rules for `[CulturedFact]` and `[CulturedTheory]`:

| ID | Severity | Reported when |
|---|---|---|
| **xUnit1070** – Cultured test methods cannot have null cultures | Error | A culture is `null`, `default`, `default(string)`, or a null `const string` |
| **xUnit1071** – Cultured test methods should not have duplicate cultures | Warning | A culture is listed more than once (`Culture '{0}' is duplicated.`) |

A null culture breaks the test at runtime. A duplicate only runs the same test twice, so it's a warning.

## What changed

- The rules are new diagnostic IDs in the same analyzer as xUnit1060, which does the same argument analysis. I renamed the class from `CulturedTestMustHaveAtLeastOneCulture` to `CulturedTestCultureValidation` because it now covers three rules. This follows `ConditionalSkipPropertyValidation` (xUnit1054/1055).
- It uses the IOperation API and handles both array creation (`new[] { ... }`) and collection expressions (`[...]`). Each bad element is reported at its own location. For duplicates, the second and later occurrences are reported.
- When the culture list is empty, only xUnit1060 is reported.

## Notes for reviewers

- **Case:** duplicates are compared ignoring case (`"en-US"` and `"en-us"` are the same culture).
- **Empty string:** `""` (invariant culture) is allowed.
- **Not covered:** invalid culture names such as `"xx-BOGUS"`. Whether a name is valid depends on the OS culture data (ICU), so the analyzer can't check it reliably.
- **xUnit1060 change:** a null cultures array (`[CulturedFact(null)]`, `default`, `(string[])null`) is now reported as xUnit1060. At runtime it fails with the same "did not provide any cultures" error as an empty array.
- **Tests:** there's one test file per ID, each with a single large `V3_only` test. Because the analyzer now reports more than one ID, the X1060 test markup names its ID (`{|xUnit1060:...|}`).
- **Docs:** the rule pages for xUnit1070 and xUnit1071 still need to be added in the xunit.net site repo.


bradwilson:
Thanks!

bradwilson:
Available in `2.2.0-pre.21` https://xunit.net/docs/using-ci-builds

